PR #13, opened 2020-07-13
Feature sessions for multiple resorts (for askers)

### LINKED ISSUE #14: View all sessions and write enquiry message for new registrations in different consulting types
**Describe the solution you'd like**
As an asker I want to register to different consulting types with only one account to get advice for different problems.
In my profile I want to have an overview over all consulting types I already registered in and be able to register to other consulting types.
In my session list (my messages) I want to be able to write an enquiry message for any new registrations and see all (new) sessions.

**Describe alternatives you've considered**
No alternatives.

**Additional context**
No additional context

koepferd:
Part of #14 

## Proposed Changes

  - refactored agencies call for new api specs & refactored agency rendering in registration
  - refactored enquiry call for new api specs
  - added writeEnquiry for detail view to be able to send another enquiry with the list view still visible 
  - refactored sessions list and list item to display a list item for new enquiries to other resorts


src/components/sessionsListItem/ts/SessionListItemComponent.tsx
```diff
@@ -182,15 +198,18 @@ export const SessionListItemComponent = (props: SessionListItemProps) => {
...
 					: `sessionsListItem`
 			}
 			data-group-id={listItem.groupId}
 		>
 			<div
 				className={
-					activeSession && activeSession.session.id === listItem.id
+					(activeSession &&
+						activeSession.session.id === listItem.id) ||
+					activeSessionGroupId === listItem.id
```
iam-robin: noticed that this is the exact same condition like the one for the sessionslistItem className in line 201.
Does is make sense to only set the active class on the sessionsListItem and react in case of the sessionsListItem__content element, when the parent has the class active?

koepferd: Thanks for spotting this 🙂

iam-robin (approved):

PR #31, opened 2020-08-26
Feature refactor login

koepferd:
## Proposed Changes

  - refactored the login from hbs to a react component
  - implemented a basic icon component


src/components/iconSet/ts/Icon.tsx
```diff
@@ -0,0 +1,61 @@
...
+	IN_VIA: 'IN_VIA',
+	KREUZBUND: 'KREUZBUND',
+	RAPHAELSWERK: 'RAPHAELSWERK',
+	SKF: 'SKF',
+	SKM: 'SKM'
+};
+
+export interface IconProps {
+	name: string;
+	className?: string;
```
iam-robin: does is make sense to define the width and height of the svg as props, too?
I'm not quite sure, because I don't know what happens when the ratio isn't the same as the viewBox.

What do you think?

koepferd: Thank you for the discussion we had. Did a bit of refactoring.

iam-robin (dismissed):

iam-robin (approved):

PR #33, opened 2020-09-03
refactor: 💡 changed forward condition, svgs refactoring

koepferd:
## Proposed Changes

  - changed condition for forward functionality availability (no forwarding on messages containing only an attachment) & tooltip text
  - refactoring of svgs (using svg component) to reduce complexity of components


src/components/message/ts/MessageItemComponent.tsx
```diff
@@ -64,6 +66,7 @@ export const MessageItemComponent = (props: MessageItemComponentProps) => {
 	const renderedMessage = contentStateMessage.hasText()
 		? urlifyLinksInText(stateToHTML(contentStateMessage))
 		: '';
+	const hasRenderedMessage = renderedMessage && renderedMessage.length > 0;
```
iam-robin, line 69: did you already check if optional chaining is polyfilled in this project?

koepferd: yes, it is not yet. Will probably do it in another refactoring task soon.

iam-robin (approved):

PR #37, opened 2020-09-21
refactor: 💡 move error page styles from exernal to inline

iam-robin:
## Proposed Changes

  - move error page styles into hbs template
  - ensure that error pages are displayed correctly even if the resources are not loaded correctly


src/components/errorPage/errorPage.hbs
```diff
@@ -33,4 +33,208 @@
...
+		margin: auto 0 auto auto;
+		display: none;
+	}
+
+	.errorPage__main {
+		background-color: var(--background-grey);
+		padding: 0 var(--grid-base-three);
+		display: flex;
+		flex-direction: column;
+		height: 100vh;
```
koepferd, line 111: please make sure to test vh on mobile devices and use different solution if necessary.

iam-robin: works on mobile devices, because we have no elements aligned to the top or bottom of the viewport.

PR #48, opened 2020-11-10
Feature liveservice direct message

koepferd:
## Proposed Changes

  - added live mySessions list reload on new directMessage
  - fixed routing rerenders
  - memoizing routes and session


package.json
```diff
@@ -61,6 +61,7 @@
         "semver": "^5.6.0",
         "sockjs-client": "1.5.0",
         "susy": "2.2.14",
+        "text-encoding": "^0.7.0",
```
iam-robin: please remove the auto-update '^' inside the package version, to prevent future problems

src/components/app/ts/app.tsx
```diff
@@ -36,32 +37,25 @@ export const AppContainer = (props) => {
...
+
+const STOMP_EVENT_TYPES = {
+	DIRECT_MESSAGE: 'directMessage'
+};
+
+export const App = () => {
 	const { setAuthData } = useContext(AuthDataContext);
 	const [authDataRequested, setAuthDataRequested] = useState(false);
-	const { userData, setUserData } = useContext(UserDataContext);
+	const { setUserData } = useContext(UserDataContext);
```
iam-robin, line 52: is there a use case to only set a state but never get or read it?

src/components/messageSubmitInterface/ts/messageSubmitInterfaceComponent.tsx
```diff
@@ -118,7 +118,7 @@ export interface MessageSubmitInterfaceComponentProps {
 	placeholder: string;
 	showMonitoringButton?: Function;
 	type: string;
-	typingUsers?: [];
+	typingUsers?: any[];
```
iam-robin: Can we find a more specific data type for the typing users?

src/components/sessionsList/ts/SessionsList.tsx
```diff
@@ -460,8 +476,10 @@ export const SessionsList = () => {
 				onScroll={handleListScroll}
 			>
 				{hasUserAuthority(AUTHORITIES.USER_DEFAULT, userData) &&
-				showNewMessageForUser ? (
-					<SessionsListUnreadIllustration />
+				sessionsData &&
```
iam-robin: please use a normal if statement, when the else case will be null.
Our new create-react-app linter will mark this as a warning 😇

src/components/app/ts/app.tsx
koepferd: yes, after login the userData is called via REST Api and saved in the global state.

iam-robin (approved):

iam-robin (approved):

PR #51, opened 2020-11-12
feat: 🎸 migrate error page to react

johannescpk:
## Proposed Changes

  - Add Error Page as React Component to CRA


src/components/error/error.styles.scss
```diff
@@ -0,0 +1,204 @@
+:root {
+	--white: #fff;
```
iam-robin: for some of the defined variables (for example colors, versions of the grid-base, etc.) you can use our global variables (resources/styles/settings.scss) instead.

iam-robin (approved):

PR #55, opened 2020-11-18
fix: 🐛 validation of confirmation pw after changing pw

iam-robin:
## Proposed Changes

  - trigger confirmation validation also on password change
  - rename parameters in passwordReset component (`pw` to `password`)
  - add typing to adjusted function parameters


iam-robin:
merge this as soon as the `Create React App` Frontend version is released.

koepferd (approved):

PR #59, opened 2020-11-20
Feature password special characters

iam-robin:
## Proposed Changes

  - change special character validation for password field (everything except of numbers (0-9) and normal characters (a-z) are now special characters)
  - adjust set password info text and add examples for common special characters.


iam-robin:
Merge this PR into develop after `feature-react-build-process` is merged into develop.

koepferd (approved):

PR #61, opened 2020-11-23
Feature cra draft message liveservice implementation

koepferd:
## Proposed Changes

  - refactor namings
  - added draftMessage saving & loading feature
  - added stompjs and liveservice implementation


src/resources/scripts/polyfills/index.ts
```diff
@@ -1,6 +1,7 @@
-import 'fetch-polyfill/fetch';
```
johannescpk, line 1: The `src/resources/scripts/polyfills` directory is no longer used. Forgot to remove that in #54. So changes here will have no effect.

src/components/app/app.tsx
```diff
@@ -1,11 +1,11 @@
-import '../../polyfill';
```
johannescpk, line 1: This would need to stay in order to have correct polyfill behavior.

johannescpk (approved):

PR #62, opened 2020-11-23
refactor(lint): use prettier with eslint and cleanup

johannescpk:
## Proposed Changes

  - Add prettier to eslint
  - Let lint-staged use eslint for js/ts files
  - Move all configurations from separate files into package.json
  - Remove unused Jest stuff and unused dependencies
  - Fix TS compiler error in `SessionsListData.tsx` (caused by package-lock being bumped)

## Notes

Configuration I'm using in VSCode to have this work fine:

- Disable Format on Save
- Have the ESLint extension installed and add the following config:

```
	"editor.codeActionsOnSave": {
		"source.fixAll.eslint": true
	},
```


johannescpk:
So it seems that nowadays Prettier no longer recommends the approach of adding it directly as plugin to ESLint:

https://prettier.io/docs/en/integrating-with-linters.html

> These plugins were especially useful when Prettier was new. By running Prettier inside your linters, you didn’t have to set up any new infrastructure and you could re-use your editor integrations for the linters. But these days you can run prettier --check . and most editors have Prettier support.
> 
> The downsides of those plugins are:
> 
> - You end up with a lot of red squiggly lines in your editor, which gets annoying. Prettier is supposed to make you forget about formatting – and not be in your face about it!
> - They are slower than running Prettier directly.
> - They’re yet one layer of indirection where things may break.

Given that it doesn't seem to make much sense to use the approach in this PR.

PR #69, opened 2020-11-25
feat: refresh keycloak access token until refresh token expires

johannescpk:
## Proposed Changes

  - refresh keycloak access token 10s before it expires
  - logout if refresh token expires or we can't refresh access token in time
  - stay logged in as long as we can refresh the access token or the access token is still valid

### Manually against BE

- Needs correct Keycloak configuration which is a bit tricky (from: https://stackoverflow.com/a/54679852)
    - http://caritas.local/auth/admin/master/console/#/realms/caritas-online-beratung/token-settings
    - SSO Session Idle: 999 days
    - SSO Session Max: 3 minutes
    - Client Session Idle: 0
    - Client Session Max: 0
    - Access Token Lifespan: 1 minute
- Log in to the app
  - Make sure that the first call to `token` in the devtools network tab responds with `expires_in: 60`, `refresh_expires_in: 180`
  - After 50, 100 and 150 seconds a new access token request should be visible in the devtools network tab
  - After 3 minutes a logout should happen

### Mocked with Cypress

Run `npm run dev` and click `tokens.spec.ts`

## Refactoring

- [x] Logging out because of expired refresh token should probably not happen as long as an access token is still valid
- [x] Only calculate time once before storing in local storage
- [x] first check in handletokenrefresh doesn't make sure access token is no longer valid?
- [x] `tokenValidMs` -> `accessTokenValidInMs`
- [x] `validUntil` -> `accessTokenValidUntilTime`
- [x] `refreshValidUntil` -> `refreshTokenValidUntilTime`

### DIFF
--- a/cypress/fixtures/auth.token.json
+++ b/cypress/fixtures/auth.token.json
@@ -1,7 +1,7 @@
 {
 	"access_token": "access_token",
-	"expires_in": 43200000,
-	"refresh_expires_in": 900,
+	"expires_in": 600,
+	"refresh_expires_in": 1800,
 	"refresh_token": "refresh_token",
 	"token_type": "bearer",
 	"not-before-policy": 0,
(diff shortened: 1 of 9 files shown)


koepferd (changes requested):
Please refactor namings and the local storage functions as discussed.

koepferd (approved):

PR #76, opened 2020-12-01
feat: show error for message attachment when quota is reached

johannescpk:
## Proposed Changes

  - Show error message when uploadservice responds with 403 and `X-Reason: "QUOTA_REACHED"` header when trying to send a message with attachment

## Testing

- On Staging: Login with an user and try to upload more than 7 files
- Locally testing: won't work because of CORS which swallows the header if not properly configured: https://stackoverflow.com/questions/8945587/xmlhttprequest-getallresponseheaders-not-returning-all-the-headers


src/resources/scripts/i18n/de/attachments.ts
```diff
@@ -4,6 +4,8 @@ const attachments = {
 		'Ihre Datei konnte nicht gesendet werden. Erlaubt sind Bilder (jpg und png), sowie Dokumente (docx und pdf). Bitte versuchen Sie es erneut.',
 	'error.size.headline': 'Ihre gewählte Datei ist zu groß.',
 	'error.size.message': 'Sie können max. 5MB hochladen.',
+	'error.quota.headline': 'Sie haben das Limit zum Hochladen erreicht.',
```
koepferd, line 7: Please also add informal texts to attachmentsInformal.ts.

koepferd (changes requested):

koepferd (approved):

koepferd:
Ready for PO Approval

PR #77, opened 2020-12-04
Feature register to group chats and vice versa

koepferd:
## Proposed Changes

  - enable new registration to consultingType 15 (groupChats) from other consultingTypes
  - enable new registration to other consultingTypes for consultingType 15 (groupChats)
  - info texts to explain other consultingType mode
  - design changes in the profile for ui consistency


src/components/agencySelection/AgencySelection.tsx
```diff
@@ -163,57 +189,70 @@ export const AgencySelection = (props: AgencySelectionProps) => {
...
-									</div>
-								) : null}
-								{agency.postcode ? (
-									<div className="askerRegistration__postcodeFlyout__postcode">
-										{agency.postcode}
+												{translate(
+													'registration.agency.prefilled.isTeam'
+												)}
+											</div>
+										) : null}
```
iam-robin: you could use the && syntax instead of the ? (...) : null condition.
Would made sense for all the conditions in this return.

iam-robin (approved):

PR #78, opened 2020-12-04
Refactor registration translations

koepferd:
## Proposed Changes

  - use translation keys for registrations to minimize redundancy
  - overline all registrations
  - other data for addiction and u25


src/components/registration/registrationData.js
```diff
@@ -1,132 +1,226 @@
...
-				headline: 'Suchtmittel',
+				headline: translate(
+					'user.userAddiction.addictiveDrugs.headline'
+				),
 				name: 'addictiveDrugs',
 				tagSelects: [
 					{
 						id: 'alcohol',
-						label: 'Alkohol'
+						label: handleNumericTranslation(
```
iam-robin: please reference the direct translate path without the helper function for readability reasons

iam-robin (approved):

PR #80, opened 2020-12-08
Feature session scroll behavior

iam-robin:


src/components/session/SessionItemComponent.tsx
```diff
@@ -161,20 +256,43 @@ export const SessionItemComponent = (props: SessionItemProps) => {
...
+				<div
+					className={`session__scrollToBottom ${
+						isScrolledToBottom
+							? 'session__scrollToBottom--disabled'
+							: ''
+					}`}
+				>
+					{newMessages > 0 && (
+						<span className="session__unreadCount">
+							{newMessages > 99 ? '99+' : newMessages}
```
koepferd: Please use a translate variable here.

src/components/session/session.styles.scss
```diff
@@ -89,6 +89,44 @@ $sessionHeaderHeight: 85px;
...
+	}
+
+	&__unreadCount {
+		display: flex;
+		justify-content: center;
+		align-items: center;
+		position: absolute;
+		top: -3px;
+		right: -9px;
+		width: 18px;
```
koepferd: You might use a variable here for lines 117 & 118

koepferd (changes requested):

src/components/session/SessionItemComponent.tsx
iam-robin: 👍

src/components/session/session.styles.scss
iam-robin: 👍

koepferd (approved):

PR #92, opened 2020-12-21
Feature complete postcode

koepferd:
## Proposed Changes

  - changes postcode validation logic for registrations
  - when postcode has to be sent it always has to be full length (5 numbers)
  - edge cases when no postcode has to be submitted
  - styling changes
  - show selected agency in ui

### DIFF
--- a/src/components/agencySelection/AgencySelection.tsx
+++ b/src/components/agencySelection/AgencySelection.tsx
@@ -39,10 +39,9 @@ export const AgencySelection = (props: AgencySelectionProps) => {
 	>(undefined);
 	const [autoSelectAgency, setAutoSelectAgency] = useState(false);
 	const [autoSelectPostcode, setAutoSelectPostcode] = useState(false);
-	const [
-		selectedAgencyData,
-		setSelectedAgencyData
-	] = useState<AgencyDataInterface | null>(null);
+	const [selectedAgencyData, setSelectedAgencyData] = useState<
+		AgencyDataInterface
+	>(null);
 
 	const validPostcode = () =>
 		selectedPostcode?.length === VALID_POSTCODE_LENGTH;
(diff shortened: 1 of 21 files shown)


src/components/agencySelection/AgencySelection.tsx
```diff
@@ -34,20 +32,37 @@ export const AgencySelection = (props: AgencySelectionProps) => {
...
-	const [typedPostcode, setTypedPostcode] = useState('');
 	const [selectedAgencyId, setSelectedAgencyId] = useState<
 		number | undefined
 	>(undefined);
 	const [autoSelectAgency, setAutoSelectAgency] = useState(false);
+	const [autoSelectPostcode, setAutoSelectPostcode] = useState(false);
+	const [
+		selectedAgencyData,
+		setSelectedAgencyData
+	] = useState<AgencyDataInterface | null>(null);
```
iam-robin: Because of our `"strictNullChecks": false` typescript configuration we do not have to implicitly define null as a type.  `useState<AgencyDataInterface>` should work here as well.

src/components/agencySelection/AgencySelection.tsx
```diff
@@ -203,6 +197,17 @@ export const AgencySelection = (props: AgencySelectionProps) => {
...
 					></InputField>
+					{!props.preselectedAgency &&
+						!autoSelectAgency &&
+						selectedAgencyData && (
+							<SelectedAgencyInfo
+								prefix={translate(
+									'registration.agency.selected.prefix'
+								)}
+								agencyData={selectedAgencyData}
+								className="selectedAgencyInfo--agencySelection"
```
iam-robin: I don't understand the BEM class name here. Is `--agencySelection` really a state of the selectedAgencyInfo?

koepferd: Done

src/components/agencySelection/AgencySelection.tsx
koepferd: Done

iam-robin (approved):

PR #98, opened 2020-12-22
feat(session-list): offset fetch when scroll threshold is reached

johannescpk:
## Proposed Changes

  - Add threshold of 50px to trigger fetching new sessions before reaching the session list end

### DIFF
--- a/cypress/integration/sessions.spec.ts
+++ b/cypress/integration/sessions.spec.ts
@@ -4,7 +4,10 @@ import {
 	sessionsReply
 } from '../support/sessions';
 import sessionListI18n from '../../src/resources/scripts/i18n/de/sessionList';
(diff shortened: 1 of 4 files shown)


cypress/integration/sessions.spec.ts
```diff
@@ -22,6 +25,74 @@ describe('Sessions', () => {
...
+		it('should fetch next batch of sessions if scroll threshold is reached', () => {
+			const amountOfSessions = 100;
+			const sessions = generateMultipleConsultantSessions(
+				amountOfSessions
+			);
+
+			cy.caritasMockedLogin({
+				type: 'consultant',
+				sessions
+			}).then(() => console.log('ho'));
```
koepferd: Please remove log/ then case.

koepferd (changes requested):

koepferd (approved):

PR #107, opened 2021-01-19
Feature incoming call styling

iam-robin:
## Proposed Changes

  - adjust incoming call styles
  - add audio to incoming call
  - rename video call translation file
  - add tooltips to video call buttons
  - add type prop to notifcation component


src/components/notifications/Notifications.tsx
```diff
@@ -1,12 +1,17 @@
...
 import { IncomingCall, IncomingCallProps } from '../incomingCall/IncomingCall';
 import './notifications.styles';
+import incomingCallRingtone from '../../resources/audio/incomingCall.mp3';
 
 interface NotificationsProps {
 	notifications: IncomingCallProps[];
 }
 
 export const Notifications = (props: NotificationsProps) => {
+	const existingCall: boolean = props.notifications.some(
```
koepferd: please rename as discussed `hasIncomingVideoCall`

koepferd (changes requested):

src/components/notifications/Notifications.tsx
iam-robin: done 👍

koepferd (approved):

PR #112, opened 2021-01-22
Hotfix Release 2021-01-25 

koepferd:
## Proposed Changes

  - changing label of caution text
  - caution text added for agency selection


src/components/infoText/InfoText.tsx
```diff
@@ -19,7 +19,7 @@ const getLabelContent = (type: string) => {
 
 	if (type === LABEL_TYPES.CAUTION) {
```
iam-robin: can we rename the label type from CAUTION to NOTE or something else?

koepferd: Yes, done: `NOTICE`

iam-robin (approved):

PR #126, opened 2021-02-11
Feature welcome registration

koepferd:
## Proposed Changes

  - added a welcome screen to every registration page
  - adjusted cypress tests for registration pages

### DIFF
--- a/src/components/inputField/inputField.styles.scss
+++ b/src/components/inputField/inputField.styles.scss
@@ -20,7 +20,7 @@
 			}
 		}
 		.formWrapper__infoText {
-			margin: 12px;
+			margin: 12px 0;
 		}
 	}
 
--- a/src/components/inputField/inputField.styles.scss
+++ b/src/components/inputField/inputField.styles.scss
@@ -20,8 +20,7 @@
 			}
 		}
 		.formWrapper__infoText {
-			margin-top: 12px;
-			margin-bottom: 12px;
+			margin: 12px;
 		}
 	}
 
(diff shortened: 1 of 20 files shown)


src/components/inputField/inputField.styles.scss
```diff
@@ -19,6 +19,10 @@
 				}
 			}
 		}
+		.formWrapper__infoText {
+			margin-top: 12px;
```
iam-robin: if the left and right values are 0 you can write: 
margin: 12px 0;
(for less redundancy)

src/components/profile/profile.styles.scss
```diff
@@ -470,7 +470,7 @@ $maxFormElementWidth: 320px;
 		}
 
 		.askerRegistration__consultingModeInfo {
-			padding: 0 $grid-base-three 0 20px;
+			padding: 0 $grid-base-three 40px 20px;
```
iam-robin: use $grid-base-five instead of 40px

src/components/inputField/inputField.styles.scss
koepferd: There was a dependency why this wasn't possible, but I fixed it and forgot to change this. Thanks for your keen eye :-)

src/components/inputField/inputField.styles.scss
```diff
@@ -20,8 +20,7 @@
 			}
 		}
 		.formWrapper__infoText {
-			margin-top: 12px;
-			margin-bottom: 12px;
+			margin: 12px;
```
iam-robin: you forgot the 0 for left and right, didn't you?

iam-robin (approved):

PR #142, opened 2021-03-22
feat: 🎸 added <LegalInformationLinks />

koepferd:
## Proposed Changes

  - added `<LegalInformationLinks />` component to `<Login />` and `<Registration />`


src/components/login/LegalInformationLinks.tsx
```diff
@@ -0,0 +1,23 @@
...
+import './legalInformationLinks.styles';
+
+export const LegalInformationLinks = () => {
+	return (
+		<div className="legalInformationLinks">
+			<Text
+				text={translate('login.legal.infoText.impressum')}
+				type={'infoSmall'}
+			/>
+			<Text
```
iam-robin, line 13: not sure if it makes sense to use the text component + translate content for the divider, as it is a graphic element and not content. Would go with an after element on each . legalInformationLinks > text (except the last item)

src/resources/scripts/i18n/de/login.ts
```diff
@@ -6,7 +6,12 @@ const login = {
 	'resetPasswort.label': 'Passwort vergessen?',
 	'register.infoText.title': 'Noch nicht registriert?',
 	'register.infoText.copy': 'Wir beraten Sie gerne zu folgenden Themen:',
-	'register.linkLabel': 'Zur Übersicht'
+	'register.linkLabel': 'Zur Übersicht',
+	'legal.infoText.impressum':
+		'<a href="https://www.caritas.de/impressum" target="_blank">Impressum</a>',
+	'legal.infoText.divider': '|',
```
iam-robin: see comment in legalInformationLinks component

iam-robin (approved):

PR #148, opened 2021-03-29
Refactor registration

amannn:
Previously we created all necessary registration pages with the `HTMLWebpackPlugin`. Now there's only a single page generated that renders the correct registration content based on the URL.

The routing is a bit tricky here, I'll add an inline comment with further explanation and to discuss this.

## Proposed Changes

  - Refactor registration to use a single HTML page that can be configured with API data


config/webpack.config.js
```diff
@@ -554,249 +553,14 @@ module.exports = function (webpackEnv) {
...
-				title: 'Registrierung Jungen- und Männerberatung',
-				templateParameters: {
-					type: 'registration',
-					consultingType: 20,
-					resortName: 'men',
-				},
-				template: 'src/pages/app.html',
-				chunks: ['registration'],
-				filename: 'registration.jungen-und-maenner.html',
+				filename: 'registration.html',
```
amannn, line 563: The routing is a bit tricky here, as we have a different one in production than in development and a mapping in between.

Previously:
```
/suchtberatung -> login.html
/suchtberatung/registration -> registration.suchtberatung.html
```

Now:
```
/suchtberatung -> login.html
/suchtberatung/registration -> registration.html?resortName=addiction
```

As we're still generating an HTML file to be able to serve the content, we need some mechanism to encode the resort in the URL. As we're generating an HTML file, a query parameter sounded like a reasonable approach. However this is slightly problematic as when we're reading the URL on the client side, we have to handle a different URL in production and in development.

PR #170, opened 2021-05-03
refactor: 💡 renaming consulting routing config

koepferd:
## Proposed Changes

  - rename consultant routing config to generic name


amannn (approved):
Nice! One more refactoring out of the way 🙂

PR #171, opened 2021-05-03
Refactor library

amannn:
Generally we have two approaches for theming and configurability:

1. Since the consumer owns the entry point, configuration can be realized with React APIs. You can for example pass custom props, a context provider or decide to render entirely different components for a given entry point.
2. Override individual files by resolving to local files in webpack (based on a provided, explicit list).

In my opinion the first option is a bit cleaner, more explicit since types can be enforced and reduces the number of integration points for the consumer. However it is only practical when we can access the config and theme via React components. To avoid bigger refactors, it's probably a good idea to use both approaches where useful. With the second option we have to be careful to provide documentation about the supported integration points, as moving files could potentially break a consumer.

Currently React APIs are proposed for entry points and allows us e.g. to customize the stage.

Local overrides on the other hand are used for:
- Translations
- SASS config
- Config like URLs

As we're still waiting for a decision for icons and illustrations, these two are not made dynamic for now. Local resolving could be an easy fix here though in the future.

Consuming repo: https://github.com/virtualidentityag/diakonie-onlineBeratung-frontend-theme/pull/1

Once these changes are in the target branch, we can create another PR against staging and test the publishing.


amannn:
Replaced by https://github.com/CaritasDeutschland/caritas-onlineBeratung-frontend/pull/172

PR #176, opened 2021-05-07
Refactor: Remove static consulting type usage

amannn:
Replaces all conditions that reference a static consulting type ID with dynamic data from the API.

Generally this seems to work. On my machine there's currently this strange issue that I can't add voluntary components after registration, could you help me with testing that case? Generally if you could help with testing this a bit that would be really good.

**TODO:**
- [ ] Wait for parent branch to be merged and change target to develop


.github/workflows/build.yml
```diff
@@ -48,3 +48,8 @@ jobs:
                   branch: ${{env.BRANCH}}
                   force: true
                   tags: true
+            - uses: actions/upload-artifact@v2
+              if: failure()
+              with:
+                  name: cypress-videos
+                  path: cypress/videos/
```
amannn, line 55: I had some issues with failing Cypress tests at some point. This can be helpful to debug the issue.

cypress/fixtures/service.users.data.askers.json
```diff
@@ -22,25 +21,7 @@
...
-		"10": { "sessionData": null, "isRegistered": false, "agency": null },
-		"11": { "sessionData": null, "isRegistered": false, "agency": null },
-		"12": { "sessionData": null, "isRegistered": false, "agency": null },
-		"13": { "sessionData": null, "isRegistered": false, "agency": null },
-		"14": { "sessionData": null, "isRegistered": false, "agency": null },
-		"15": { "sessionData": null, "isRegistered": false, "agency": null },
-		"16": { "sessionData": null, "isRegistered": false, "agency": null },
-		"17": { "sessionData": null, "isRegistered": false, "agency": null },
-		"18": { "sessionData": null, "isRegistered": false, "agency": null },
-		"19": { "sessionData": null, "isRegistered": false, "agency": null }
```
amannn, line 43: Most of these don't exist in our consulting type mocks.

cypress/integration/tokens.spec.ts
```diff
@@ -11,10 +11,23 @@ const waitForTokenProcessing = () => {
 
 describe('Keycloak Tokens', () => {
 	let authTokenJson;
-	before(() => {
+	beforeEach(() => {
```
amannn, line 14: Interestingly this change was necessary to get the tests to pass. I don't know why that would make a difference but now tests work consistently. Also this was only an issue on CI.

src/components/app/AuthenticatedApp.tsx
```diff
@@ -85,14 +86,15 @@ export const App: React.FC = () => {
 	if (!userDataRequested) {
 		setUserDataRequested(true);
 		handleTokenRefresh().then(() => {
-			apiGetUserData()
-				.then((userProfileData: UserDataInterface) => {
+			Promise.all([apiGetUserData(), apiGetConsultingTypes()])
```
amannn, line 89: This is an important change. Since we need to access consulting type data in quite a few places in the authenticated part of the app, I thought it might be a good idea to fetch them all at first and therefore being able to use them synchronously afterwards.

Alternatively we'd have much more asynchronicity along with loading and error states spread across components and would have to find a solution with caching.

src/components/askerInfo/AskerInfoData.tsx
```diff
@@ -3,8 +3,7 @@ import { useContext } from 'react';
 import {
 	translate,
 	handleNumericTranslation,
-	getAddictiveDrugsString,
-	getResortTranslation
+	getAddictiveDrugsString
```
amannn, line 6: As discussed we'll leave this for the time being.

src/components/askerInfo/AskerInfoData.tsx
```diff
@@ -18,13 +17,15 @@ import {
...
 
 export const AskerInfoData = () => {
 	const { sessionsData } = useContext(SessionsDataContext);
 	const { activeSessionGroupId } = useContext(ActiveSessionGroupIdContext);
 	const activeSession = getActiveSession(activeSessionGroupId, sessionsData);
-
-	const resort = getResortTranslation(activeSession.session.consultingType);
+	const consultingType = useConsultingType(
+		activeSession.session.consultingType
+	);
```
amannn, line 28: This hook makes it convenient to read a single consulting type during render and takes care of error handling.

src/components/profile/AskerAboutMeData.tsx
```diff
@@ -20,24 +21,14 @@ export const AskerAboutMeData = () => {
...
-		hasAskerEmailFeatures(userData).then((result) => {
-			if (isCanceled) return;
-			setShowEmail(result);
-		});
-		return () => {
-			isCanceled = true;
-		};
-	}, [userData]);
+	const { consultingTypes } = useContext(ConsultingTypesContext);
+	const showEmail = hasAskerEmailFeatures(userData, consultingTypes);
```
amannn, line 30: Was able to turn this into a synchronous call now that we have consulting types generally available.

src/components/profile/profileHelpers.ts
```diff
@@ -55,25 +57,32 @@ export const getConsultingTypesForRegistrationStatus = (
 			const validationForRegistrationStatus =
 				registrationStatus === REGISTRATION_STATUS_KEYS.REGISTERED
 					? value.data.isRegistered
-					: !forAskerRegistrationExcludedConsultingTypes.includes(
-							parseInt(value.consultingType)
-					  ) && !value.data.isRegistered;
+					: consultingTypes.find(
+							(cur) => cur.id === parseInt(value.consultingType)
+					  )?.isSubsequentRegistrationAllowed &&
```
amannn, line 62: The flag here is inverted, hence the negation is gone.

src/globalState/interfaces/ConsultingTypeInterface.ts
```diff
@@ -13,21 +13,35 @@ export type RegistrationNotesInterface = {
...
+	groupChat: {
+		isGroupChat: boolean;
+		groupChatRules: [string];
+	};
+}
+
+export interface ConsultingTypeInterface extends ConsultingTypeBasicInterface {
+	requiredComponents?: RequiredComponentsInterface;
+	languageFormal: boolean;
+	voluntaryComponents?: any[];
```
amannn, line 48: Split into two types now. The upper one is the one coming from the list endpoint, the other one from the detail endpoint.

src/components/sessionHeader/SessionHeaderComponent.tsx
```diff
@@ -216,7 +216,7 @@ export const SessionHeaderComponent = (props: SessionHeaderProps) => {
 				<div
 					className={
 						hasUserAuthority(AUTHORITIES.ASKER_DEFAULT, userData) ||
-						isGenericConsultingType(chatItem.consultingType)
+						!consultingType.showAskerProfile
```
amannn, line 219: Here we have to negate the flag, right? Below there's the same story.

src/components/profile/AskerRegistration.tsx
```diff
@@ -101,19 +98,17 @@ export const AskerRegistration = () => {
...
 				.then((response) => {
 					let overlayItem = overlayItemNewRegistrationSuccess;
-					if (isGroupChatConsultingType(selectedConsultingType)) {
+					if (selectedConsultingType?.groupChat.isGroupChat) {
 						overlayItem.buttonSet[0].label = translate(
 							'profile.data.registerSuccess.overlay.button1Label.groupChats'
 						);
-					} else if (
-						!isGroupChatConsultingType(selectedConsultingType)
-					) {
```
amannn, line 116: Not sure why the condition was necessary here? It's just the opposite of the condition in `if`, right?

src/components/agencySelection/AgencySelection.tsx
```diff
@@ -166,20 +155,15 @@ export const AgencySelection = (props: AgencySelectionProps) => {
 		content: selectedPostcode,
 		maxLength: VALID_POSTCODE_LENGTH,
 		pattern: '^[0-9]+$',
-		disabled:
-			autoSelectPostcode ||
-			(!props.selectedConsultingType &&
-				props.selectedConsultingType !== 0),
```
amannn, line 172: I didn't fully understand this condition here. Should there be special behaviour for the consulting type with the ID `0`?

src/components/agencySelection/AgencySelection.tsx
```diff
@@ -131,12 +121,11 @@ export const AgencySelection = (props: AgencySelectionProps) => {
 					.catch((error) => {
 						if (
 							error.message === FETCH_ERRORS.EMPTY &&
-							props.selectedConsultingType
+							props.consultingType.id
```
amannn: Should this be:

```suggestion
							props.consultingType.id != null
```

? Or are we purposefully omitting the consulting type `0` here? That would mean we have hardcoded access to it though.

src/components/profile/AskerRegistration.tsx
```diff
@@ -33,28 +33,28 @@ import { AgencySelection } from '../agencySelection/AgencySelection';
...
 	const [overlayActive, setOverlayActive] = useState(false);
 	const [overlayItem, setOverlayItem] = useState<OverlayItem>(null);
 	const { setAcceptedGroupId } = useContext(AcceptedGroupIdContext);
 	const [isRequestInProgress, setIsRequestInProgress] = useState(false);
+	const { consultingTypes } = useContext(ConsultingTypesContext);
+	const selectedConsultingType = useConsultingType(selectedConsultingTypeId);
 
-	const isAllRequiredDataSet = () => selectedConsultingType && selectedAgency;
+	const isAllRequiredDataSet = () =>
+		selectedConsultingTypeId && selectedAgency;
```
amannn: Should this be:

```suggestion
		selectedConsultingTypeId != null && selectedAgency;
```

? Or is there special behaviour for the consulting type `0`?

src/components/profile/profileHelpers.ts
```diff
@@ -55,25 +57,32 @@ export const getConsultingTypesForRegistrationStatus = (
...
-		const currentConsultingType = parseInt(value.consultingType);
+		const id = parseInt(value.consultingType);
+		const consultingType = getConsultingType(consultingTypes, id);
+
 		return {
 			value: value.consultingType,
-			label: isGroupChatConsultingType(currentConsultingType)
-				? getResortTranslation(currentConsultingType, false, true)
-				: getResortTranslation(currentConsultingType)
+			label: consultingType.groupChat.isGroupChat
```
koepferd: Let's refactor this as discussed. The title.registrationDropdown will be required in the consultingType Data and can always be used here.

src/globalState/interfaces/ConsultingTypeInterface.ts
koepferd: Instead of adding a new interface I'd suggest to implement the interface via `dtsgen`. Though I think we might need until the consultingTypeService is available for open source.

src/globalState/state.tsx
```diff
@@ -11,6 +11,7 @@ import {
 	UnreadSessionsStatusProvider,
 	UserDataProvider
 } from '.';
+import { ConsultingTypesProvider } from './provider/ConsultingTypesProvider';
```
koepferd: Please import from the `index.ts` to stay consistent here.

cypress/fixtures/service.users.data.askers.json
koepferd: Okay but please keep ct `0` and `17` as we have mocked data for these cts.

cypress/integration/tokens.spec.ts
koepferd: Hm, I don't know either why that should make a difference.

src/components/sessionHeader/SessionHeaderComponent.tsx
koepferd: Correct 👌🏻

koepferd, line 219: As I tested this: useConsultingType does not include showAskerProfile.

src/components/agencySelection/AgencySelection.tsx
```diff
@@ -166,20 +155,15 @@ export const AgencySelection = (props: AgencySelectionProps) => {
...
 	};
 
 	const handlePostcodeInput = (e) => {
 		setSelectedPostcode(e.target.value);
 	};
 
-	const showPreselectedAgency =
-		preselectedAgency &&
-		!autoselectPostcodeForConsultingType(props.selectedConsultingType);
+	const showPreselectedAgency = preselectedAgency && autoSelectPostcode;
```
koepferd: This might cause the bug I told you about:
`const showPreselectedAgency = preselectedAgency && !autoSelectPostcode;`

The agency should be shown when the postcode is not autoselected.

src/components/agencySelection/AgencySelection.tsx
koepferd: Hm, I think this might actually an error. 

Anyway it's deprecated as by now the input field is only displayed when you already selected a consulting type and `autoSelectPostcode` is disabled -> please remove the disabled property here.

src/components/profile/AskerRegistration.tsx
koepferd: Nope, you are correct, please commit this change!

src/components/agencySelection/AgencySelection.tsx
```diff
@@ -57,20 +57,14 @@ export const AgencySelection = (props: AgencySelectionProps) => {
...
 		setProposedAgencies(null);
-		setAutoSelectAgency(
-			autoselectAgencyForConsultingType(props.selectedConsultingType)
-		);
-		setAutoSelectPostcode(
-			autoselectPostcodeForConsultingType(props.selectedConsultingType)
-		);
 		setPreselectedAgency(props.preselectedAgency);
-	}, [props.selectedConsultingType, props.preselectedAgency]);
+	}, [props.preselectedAgency]);
```
koepferd: This useEffect should also reset when `props. consultingType` changes.

src/components/agencySelection/AgencySelection.tsx
koepferd: Yes, please commit your suggestion :-)

koepferd (changes requested):

src/components/sessionHeader/SessionHeaderComponent.tsx
koepferd: Please check and if needed suggest a change to the api here.

src/globalState/interfaces/ConsultingTypeInterface.ts
amannn: As discussed last week, I made a note in DOB-83 to take care of this once the consulting type service is open sourced and available for the Caritas. I also left a comment in this file.

src/components/profile/profileHelpers.ts
amannn: I've removed the condition to always use `registrationDropdown` and added a note in DOB-10 so we consider this when setting up the consulting types.

src/components/agencySelection/AgencySelection.tsx
amannn: Well spotted, thanks!

amannn: It seems like the bug we discussed was caused by another wrong condition though. I've got it fixed now.

src/components/sessionHeader/SessionHeaderComponent.tsx
amannn: This is now available in `docker.pkg.github.com/virtualidentityag/diakonie-onlineberatung-consultingtypeservice/diakonie-onlineberatung-consultingtypeservice:dockerimage.v.5.feat-extension-of-basic-response-with-show-asker-profile`.

koepferd (approved):

koepferd:
Open Todo:

- [ ] await testing with configured consulting types

PR #192, opened 2021-05-25
feat: configurable stage

amannn:
Please see inline comments

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -7,4 +7,5 @@
 export { App } from './src/components/app/app';
 export { Login } from './src/components/login/Login';
 export { Error } from './src/components/error/Error';
+export { translate } from './src/utils/translate';
 export * from './src/globalState';
--- a/index.ts
+++ b/index.ts
@@ -4,8 +4,17 @@
  * but reuse as much code from this repository as useful.
  */
 
+// Page components
 export { App } from './src/components/app/app';
 export { Login } from './src/components/login/Login';
 export { Error } from './src/components/error/Error';
+
+// Component library
+export { Headline } from './src/components/headline/Headline';
+export { Text } from './src/components/text/Text';
+
+// Utils
 export { translate } from './src/utils/translate';
+
+// Types
 export * from './src/globalState';
--- a/src/components/app/LoginLoader.tsx
+++ b/src/components/app/LoginLoader.tsx
@@ -1,14 +1,19 @@
 import * as React from 'react';
-import { useEffect, useState } from 'react';
+import { ComponentType, useEffect, useState } from 'react';
 import { useParams } from 'react-router-dom';
 import { apiGetConsultingType } from '../../api';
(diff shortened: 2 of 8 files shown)


src/components/registration/Registration.tsx
```diff
@@ -13,14 +12,11 @@ import { apiGetConsultingType } from '../../api';
 import '../../resources/styles/styles';
 import './registration.styles';
 
-export const initRegistration = () => {
-	ReactDOM.render(
-		<Registration />,
-		document.getElementById('registrationRoot')
-	);
-};
```
amannn, line 21: This was a leftover and isn't used anymore, since we render the registration within the app entry point.

index.ts
```diff
@@ -4,7 +4,17 @@
...
 export { App } from './src/components/app/app';
 export { Login } from './src/components/login/Login';
 export { Error } from './src/components/error/Error';
+
+// Component library
+export { Headline } from './src/components/headline/Headline';
+export { Text } from './src/components/text/Text';
+
+// Utils
+export { translate } from './src/utils/translate';
```
amannn, line 17: I exported some more generic modules to be used in a themed repo.

src/components/app/LoginLoader.tsx
```diff
@@ -1,14 +1,19 @@
...
-import { useEffect, useState } from 'react';
+import { ComponentType, useEffect, useState } from 'react';
 import { useParams } from 'react-router-dom';
 import { apiGetConsultingType } from '../../api';
 import { Login } from '../login/Login';
+import { StageProps } from '../stage/stage';
 
 export interface LoginLoaderProps {
 	handleUnmatch: () => void;
+	stageComponent: ComponentType<StageProps>;
```
amannn, line 10: The stage is now configurable via props. It's only two entry points where we need to pass it, therefore I thought just passing it via props might be the simplest solution for now.

src/components/stage/stage.styles.scss
```diff
@@ -92,8 +92,11 @@ $gridSpacing: $grid-base-four;
 	&--animated {
 		display: flex;
 		width: 100vw;
-		animation: overlayAnimation 1.4s ease forwards;
-		animation-delay: $loginOverlayDelayMobile;
+
+		&.stage--ready {
+			animation: overlayAnimation 1.4s ease forwards;
+			animation-delay: $loginOverlayDelayMobile;
+		}
```
amannn, line 99: Seems like the `--ready` styles got lost while resolving a merge conflict. This fixes the behaviour. Previously the stage might open too soon, while data is still being loaded.

src/components/registration/Registration.tsx
koepferd: Yeah - we already removed it in another feature aswell 😀

koepferd (approved):

PR #193, opened 2021-05-26
feat: provide theming for components

amannn:
Changes:

- Provides new SASS variables so the Diakonie app can theme the components as necessary.
- A few small general improvements that are relevant both for the Diakonie as well as Caritas.

Please see inline comments.

TODO:
- [x] Change target branch to develop once parent is merged


src/components/app/authenticatedApp.styles.scss
```diff
@@ -59,14 +59,14 @@ a {
 		flex-direction: row;
 		position: relative;
 		overflow: hidden;
-		background-color: $background-ochre;
+		background-color: $background-light;
```
amannn, line 62: I've renamed this to something more generic that doesn't imply a specific color value.

src/components/app/navigation.styles.scss
```diff
@@ -49,8 +49,6 @@ $unreadBubbleAnimation: count 500ms ease 500ms forwards;
 		width: $grid-base-five;
 		margin: 0;
 		color: $white;
-		opacity: 0.7;
-		fill-opacity: 0.7;
```
amannn, line 53: The opacity for the title and icon in the navigation was generally a bit fine-tuned.

src/components/app/navigation.styles.scss
```diff
@@ -161,6 +171,8 @@ $unreadBubbleAnimation: count 500ms ease 500ms forwards;
 		line-height: 16px;
 		text-align: center;
 		margin: auto;
+		opacity: 0.6;
+		letter-spacing: 0.5px;
```
amannn, line 175: For better readability.

src/components/button/button.styles.scss
```diff
@@ -15,7 +14,7 @@ $buttonBoxShadow: 0 6px 0 0 rgba(0, 0, 0, 0.1);
 		max-width: $buttonMaxWidth;
 		height: $buttonHeight;
 		padding: 0 $grid-base-three;
-		border-radius: $grid-base-three;
+		border-radius: $button-border-radius;
```
amannn, line 17: I've added a few variables here for theming.

src/components/button/button.styles.scss
```diff
@@ -114,8 +117,9 @@ $buttonBoxShadow: 0 6px 0 0 rgba(0, 0, 0, 0.1);
 			margin-right: $grid-base;
 		}
 
-		&:hover {
+		&:not([disabled]):hover {
 			color: $hover-primary;
+			text-decoration: underline;
```
amannn: This affects the Caritas too. I hope it's ok as a general improvement. Otherwise we can of course add another variable.

src/components/button/button.styles.scss
```diff
@@ -138,9 +142,19 @@ $buttonBoxShadow: 0 6px 0 0 rgba(0, 0, 0, 0.1);
...
+		border: 1px solid $button-small-icon-background-color-default;
 		transition: all 0.6s ease;
 
+		&:hover {
+			background-color: $white;
+
+			svg {
+				fill: $button-small-icon-background-color-default;
+			}
+		}
```
amannn, line 155: I've added a hover state for the generic "small icon" button – the other ones keep their existing hover state. The scroll to bottom button in the session view is the only call site for this variant.

src/components/formAccordion/formAccordionItem.styles.scss
```diff
@@ -57,11 +57,17 @@
...
 		}
 	}
 
 	&__continueButton {
 		margin-top: $grid-base-three;
+
+		button {
+			min-width: 0;
+			padding: 0;
+		}
```
amannn, line 70: Improvements for Diakonie. Caritas still looks the same.

src/components/login/login.styles.scss
```diff
@@ -10,15 +10,15 @@ $spacingWithIcon: 52px;
...
 		flex-direction: column;
 		padding: 0 $grid-base-two;
-		text-align: center;
+		text-align: $login-text-align;
 
 		@include breakpoint($fromLarge) {
 			width: 60vw;
 			left: 40vw;
-			padding: 0 $grid-base-three;
+			padding: 0 150px;
```
amannn, line 21: Some style improvements for the Diakonie here as well which don't affect the look of Caritas.

src/components/message/message.styles.scss
```diff
@@ -247,7 +247,7 @@ $message-lineheight: 21px;
 
 		&--myMessage {
 			color: $white;
-			background-color: $message-bg-red;
+			background-color: $message-background;
```
amannn, line 250: This variable is more generic now as well to not imply a specific color value.

src/components/monitoring/monitoring.styles.scss
```diff
@@ -108,7 +108,7 @@
 					position: absolute;
 					width: calc(100% + 32px);
 					height: 100%;
-					background-color: rgba(204, 30, 28, 0.06);
+					background-color: rgba($primary, 0.05);
```
amannn, line 111: Removed the hardcoded color value in favour of variable usage.

src/components/overlay/Overlay.tsx
```diff
@@ -22,6 +23,7 @@ export const OVERLAY_FUNCTIONS = {
 export const OVERLAY_RESET_TIME = 10000;
 
 export interface OverlayItem {
+	className?: string;
```
amannn: I'm a bit unsure what goes on the `item` part and what on `props` directly. How does this pattern work?

src/components/profile/DeleteAccount.tsx
```diff
@@ -48,6 +48,7 @@ export const DeleteAccount = () => {
 	};
 
 	const overlayConfirm: OverlayItem = {
+		className: 'deleteAccount__overlay',
```
amannn: I think there was a bug here previously:

Screenshot 2021-05-26 at 14 40 28

The delete account styles moved the text to be left-aligned, however this didn't affect just the delete account overlay, but all overlays.

Screenshot 2021-05-26 at 14 40 34


I've now scoped the styles to this component.

src/components/registration/registration.styles.scss
```diff
@@ -19,8 +17,10 @@ $scrollbarWidth: 20px;
 		}
 
 		@include breakpoint($fromLarge) {
-			width: calc(60vw - #{$scrollbarWidth} * 2);
-			left: calc(40vw + #{$scrollbarWidth});
+			width: 60vw;
+			left: 40vw;
+			padding: 0 150px;
+			max-width: 1000px;
```
amannn, line 23: An improvement that is helpful for the Diakonie but, again, doesn't affect the Caritas.

src/components/select/SelectDropdown.tsx
```diff
@@ -56,11 +56,14 @@ const colourStyles = {
...
-	option: (styles, { isFocused }) => {
+	option: (styles) => {
 		return {
 			...styles,
-			backgroundColor: isFocused ? '#F8DEDD' : '#fff',
-			color: isFocused ? '#CC1E1C' : '#3F373F',
+
+			// Use values from stylesheet
+			color: undefined,
+			backgroundColor: undefined,
```
amannn, line 65: By using `undefined`, the inline styles are not used but we can instead use values from the style sheet.

src/components/select/select.react.styles.scss
```diff
@@ -34,6 +34,20 @@ $inputHeight: 50px;
...
+
+		&__option {
+			background-color: $white;
+			color: $secondary;
+
+			&--is-focused {
+				background-color: $background-accent;
+				color: $primary;
+			}
+		}
```
amannn, line 50: Added styles for what was removed from the inline styles, so we can use SASS variables for theming.

src/components/session/SessionItemComponent.tsx
```diff
@@ -395,6 +401,5 @@ const monitoringButtonItem: ButtonItem = {
 
 const scrollBottomButtonItem: ButtonItem = {
 	icon: <ArrowDoubleDownIcon />,
-	type: BUTTON_TYPES.SMALL_ICON,
-	smallIconBackgroundColor: 'grey'
+	type: BUTTON_TYPES.SMALL_ICON
```
amannn, line 404: This call site was a bit tricky. The color of the small icon button is generally configured via props. However our theming only works with SASS variables currently. We could theoretically add theming via React context, but as long as we can stick to only SASS variables, I think we should do that first.

The trick here is that the grey variant is also the default of the button. Now the Diakonie can just style the default color of the button and therefore achieve the blue in their theme, while we can have the grey one for Caritas.

src/components/session/session.styles.scss
```diff
@@ -89,6 +90,14 @@ $unreadButtonSize: 18px;
 		}
 	}
 
+	&__submit-interface {
+		transition: box-shadow 0.3s ease;
+
+		&--scrolled-up {
+			box-shadow: 0 -2px 8px 0 rgba($black, 0.1);
+		}
+	}
```
amannn, line 99: Small general improvement: There's now a shadow when the user scrolled up in the session.

src/resources/styles/settings.scss
```diff
@@ -31,10 +31,8 @@ $background-accent: #f8dedd;
 $hover-primary: #a31816;
 $hover-select: #f4d2d1;
 
-$background-dark: #007c87;
-$background-light: #03afbf;
-$background-ochre: #f4f0ee;
-$background-light-ochre: #faf6f3;
```
amannn, line 37: `background-dark` and `background-light` were not used. I've removed them and instead renamed `ochre` to generic background variables so they can be themed without dictating a specific color value.

src/resources/styles/settings.scss
```diff
@@ -45,10 +43,6 @@ $form-error: #ff0000;
 $form-disabled: #c4bfc4;
 $form-active: #3f373f;
 
-$light-success: rgba(79, 204, 92, 0.12);
-$light-attention: rgba(255, 159, 0, 0.12);
-$light-error: rgba(255, 0, 0, 0.08);
-$light-neutral: rgba(53, 58, 74, 0.1);
```
amannn, line 51: Some of these were not used. I've renamed `light-success` and `light-neutral` to "illustration" color variables below, since that is what they are being used for.

src/resources/styles/settings.scss
```diff
@@ -58,16 +52,6 @@ $dark-brown: #432000;
...
 
-$datepicker-hover-color: $background-accent;
-$form-accordion-index-background-active: $background-accent;
-$form-accordion-index-border: 2px solid $tertiary;
-$form-accordion-index-color-active: $secondary;
-$form-accordion-index-color: $tertiary;
-$form-input-border: #dcd9dc;
-$message-bg-red: #d64b49;
-$upload-progress: #80dd92;
-$welcome-screen-icon-background: $background-accent;
```
amannn, line 69: I've moved these component variables below to the other ones.

src/resources/styles/settings.scss
```diff
@@ -129,12 +113,6 @@ $large: 1280px;
 $xlarge: 1350px;
 $medium: 1050px;
 
-// COMPONENT STYLES
-$content-max-width: $xlarge;
-$component-max-width: 1200px;
-$max-input-width: 320px;
-$sendInputHeight: 48px;
```
amannn, line 136: Also these were moved to the components section below. Keeping them at the bottom enables us to use all previously defined variables.

src/components/profile/DeleteAccount.tsx
koepferd: yeah, thanks, already fixed in the anonymous chat feature 👏🏻

src/components/overlay/Overlay.tsx
koepferd: This pattern is deprecated - I'd prefer it to only be in the `props`. So please add the `className` there. I will make an additional refactoring to change all overlays these days.

src/components/button/button.styles.scss
```diff
@@ -162,39 +176,59 @@ $buttonBoxShadow: 0 6px 0 0 rgba(0, 0, 0, 0.1);
...
+				svg {
+					fill: $white;
+				}
 			}
 		}
 
 		&--yellow {
 			color: $secondary;
 			background-color: $yellow-3;
+			border: none;
```
koepferd, line 219: This shouldn't be border `none`:
image
image

src/components/button/button.styles.scss
koepferd: I don't mind it as a general improvement as long as Eileen is fine with this from the design perspective.

src/components/session/SessionItemComponent.tsx
koepferd: nice, good solution :-)

koepferd (changes requested):

src/components/button/button.styles.scss
amannn: Hmm, how were you able to trigger the below state?

To me the button looks quite identical on `develop` and in this branch:

Screenshot 2021-06-04 at 09 25 04
Screenshot 2021-06-04 at 09 25 46

Seems like I'm missing something?

src/components/button/button.styles.scss
amannn: I just talked to Eileen about this style. It seems like this was a mistake and so it's removed both for Caritas and the Diakonie now.

src/components/button/button.styles.scss
koepferd: I tried to re-create the issue and couldn't, all good.

koepferd (approved):

PR #194, opened 2021-05-26
Feature cobh 2959 v1

menelaos:
Fixes COBH-3723, COBH-3724, COBH-3725


src/components/sessionHeader/SessionHeaderComponent.tsx
```diff
@@ -216,7 +219,8 @@ export const SessionHeaderComponent = (props: SessionHeaderProps) => {
 				<div
 					className={
 						hasUserAuthority(AUTHORITIES.ASKER_DEFAULT, userData) ||
-						isGenericConsultingType(chatItem.consultingType)
+						isGenericConsultingType(chatItem.consultingType) ||
+						isLiveChat
```
iam-robin, line 223: for better readability, you can use the clsx library to implement the conditional class name.

src/globalState/helpers/stateHelpers.ts
```diff
@@ -142,3 +143,9 @@ export const getSessionsDataWithChangedValue = (
 
 	return sesData;
 };
+
+export const isAnonymousSession = (
+	session: SessionItemInterface | null | undefined
```
iam-robin, line 148: since we have set `strictNullChecks` to `false` in the typescript config, null is always a valid value and does not need to be explicitly added as a type.

iam-robin:
I have also noticed that many SVG files in the resource/img/icon folder are displayed as a change with the message "Diff not rendered" and "Empty file". Do you know what the reason for this is?

menelaos:
Closing in favour of https://github.com/CaritasDeutschland/caritas-onlineBeratung-frontend/pull/195

PR #197, opened 2021-05-30
Feature cobh 3720 v1

menelaos:
Fixes COBH-3720

### DIFF
--- a/src/components/session/SessionItemComponent.tsx
+++ b/src/components/session/SessionItemComponent.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import { useState, useContext, useEffect, useMemo } from 'react';
 import {
+	SESSION_LIST_TAB,
 	typeIsSession,
 	typeIsTeamSession,
 	typeIsEnquiry,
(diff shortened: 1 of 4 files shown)


src/components/session/SessionItemComponent.tsx
```diff
@@ -403,6 +430,16 @@ export const SessionItemComponent = (props: SessionItemProps) => {
 					/>
 				</OverlayWrapper>
 			)}
+			{enquiryTakenByOtherConsultantOverlayActive && (
```
iam-robin, line 433: If we need more than one overlay in a component, we use to have only one overlay structure in the render and store the respective overlay item in a state, as they can never both be displayed at the same time.

You can look at the implementation e.g. in AskerRegistration.tsx

src/resources/scripts/i18n/de/session.ts
```diff
@@ -2,6 +2,9 @@ const session = {
 	'acceptance.overlayHeadline':
 		'Sie haben die Erstanfrage erfolgreich angenommen und finden diese nun unter „Meine Nachrichten“.',
 	'acceptance.buttonLabel': 'Antworten',
+	'anonymous.takenByOtherConsultant.overlayHeadline':
+		'Diese Erstanfrage wurde bereits von einem anderen Berater angenommen.',
+	'anonymous.takenByOtherConsultant.buttonLabel': 'Schliessen',
```
iam-robin, line 7: you can use 'ß' here, as the font automatically makes an 'ss' out of it in the case of uppercase

menelaos:
Closing in favour of https://github.com/CaritasDeutschland/caritas-onlineBeratung-frontend/pull/198

PR #199, opened 2021-06-01
feat: improve error page styling and use translations

amannn:
Updates the styles of the error page based on the latest design.

TODO:
- [x] Update target branch when the base is merged.


src/components/error/Error.tsx
```diff
@@ -1,71 +1,75 @@
...
+import React from 'react';
+import clsx from 'clsx';
 import { ReactComponent as Icon401 } from '../../resources/img/illustrations/kein-zutritt.svg';
 import { ReactComponent as Icon404 } from '../../resources/img/illustrations/ooh.svg';
 import { ReactComponent as Icon500 } from '../../resources/img/illustrations/gleich-zurueck.svg';
+import { translate } from '../../utils/translate';
+import { Button, BUTTON_TYPES } from '../button/Button';
 import '../../resources/styles/styles';
 import './error.styles';
-import '../button/button.styles';
```
amannn, line 9: Previously the button styles were imported without using the actual button component, but just the class names.

Somehow with the changes I made I ended up with a webpack error during the build that claimed there was a conflicting order with the stylesheets (see https://github.com/facebook/create-react-app/issues/5372). By using the full button component as well as changing the order of the styles in the login component I was able to resolve this issue.

src/components/error/Error.tsx
```diff
@@ -1,71 +1,75 @@
...
+	}
 
 	return (
-		<div className="errorPage">
-			<nav className="errorPage__nav"></nav>
+		<div className={clsx('errorPage', `errorPage--${type}`)}>
 			<header className="errorPage__header">
 				<div className="errorPage__headerMobile">
-					<h2>Beratung & Hilfe</h2>
+					<h2>{translate('error.title')}</h2>
```
amannn: Now all strings in this component are retrieved through the translation system. We don't actually need to overwrite any of these for the Diakonie, but I figured it might be preferable anyway.

koepferd: Yes, thank you :-)

koepferd: Please use the existing translations for `error.title` and `error.claim` as they should always be consistent to the translations in the general app.

koepferd (changes requested):

src/components/error/Error.tsx
amannn: Good idea. I've adjusted the identifier and removed the duplicate labels.

koepferd (approved):

PR #217, opened 2021-06-18
feat: 🎸 update monitoring keys for u25/gse

koepferd:
## Proposed Changes

  - updating the monitoring keys for u25 and gse consulting types
  - keeping the now deprecated values as they are kept in the users database and at one point might be displayed again


src/resources/scripts/i18n/de/monitoringU25.ts
```diff
@@ -86,20 +83,57 @@ const monitoringU25 = {
...
 
+	// DEV-NOTE: these keys are currently not in use but should be kept as they are still in the database and might be displayed again
+	'monitoringU25.schoolStudies': 'Schule/Studium',
+
+	'monitoringU25.partnerLivingCommunity': 'mit Partner_in in WG',
+
+	'monitoringU25.concern': 'Sorge um Suizidgefährdete_n',
+	'monitoringU25.mourning': 'Trauer nach Suizid',
+
+	// 'monitoringU25.eatingDisorder': 'Essstörung',
```
iam-robin: remove comment, eatingDisorder is already defined above

iam-robin (approved):

PR #219, opened 2021-06-23
feat: support external agencies in profile registration

amannn:
## Proposed Changes

  - With the upcoming agency service changes there can be external agencies where we need to forward to a landing page instead of handling the registration internally

## TODO

- [ ] When this is merged, publish a new version of the library for the Diakonie


cypress/integration/profile.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+
+		cy.caritasMockedLogin();
+		cy.contains('Profil').click();
+
+		cy.get('#consultingTypeSelect').click();
+		cy.get('.select__input__option:contains("Suchtberatung")').click();
+
+		cy.get('[name="postcode"]').type('00000');
+		cy.contains('Schwangerschaftsberatung Baden');
+		cy.contains('Registrieren').click();
```
amannn, line 53: TBD: Potentially we'd want to show a hint here.

src/components/agencySelection/AgencySelection.tsx
```diff
@@ -83,6 +83,7 @@ export const AgencySelection = (props: AgencySelectionProps) => {
 	useEffect(() => {
 		if (isSelectedAgencyValidated()) {
 			const agency = {
+				...proposedAgencies.find((cur) => cur.id === selectedAgencyId),
```
amannn: We need the `external` and `url` property.

src/components/profile/AskerRegistrationExternalAgencyOverlay.tsx
```diff
@@ -0,0 +1,61 @@
...
+								'profile.externalRegistration.cancel'
+							),
+							function: OVERLAY_FUNCTIONS.CLOSE,
+							type: BUTTON_TYPES.SECONDARY
+						}
+					]
+				}}
+				handleOverlay={handleOverlay}
+			/>
+		</OverlayWrapper>
```
amannn, line 58: Maybe there will be some design changes here, I've asked Eileen if we can reuse some default styles (see Teams group chat).

src/resources/scripts/i18n/de/profile.ts
```diff
@@ -52,6 +52,13 @@ const profile = {
 	'data.registerError.overlay.headline':
 		'Es ist ein Fehler aufgetreten. Bitte versuchen Sie es erneut.',
 	'data.registerError.overlay.buttonLabel': 'Schliessen',
+	'externalRegistration.headline':
+		'Ihre gewählte Beratungsstelle nutzt eine andere Anwendung für die Beratung',
+	'externalRegistration.copy.start': 'Möchten Sie für „',
+	'externalRegistration.copy.end':
+		'“ zu der anderen Anwendung wechseln und sich dort registrieren? Ihre bisherigen Beratungs- und Hilfethemen finden Sie weiterhin hier.',
```
amannn, line 59: Since I need to interpolate a consulting type title into the label, is it the right way to split the label into two?

src/components/profile/AskerRegistrationExternalAgencyOverlay.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+const AskerRegistrationExternalAgencyOverlay = ({
+	consultingType,
+	handleOverlayAction
+}: AskerRegistrationExternalAgencyOverlayProps) => {
+	const handleOverlay = (action) => {
+		if (action === OVERLAY_FUNCTIONS.REDIRECT) {
+			window.location.href =
+				consultingType.urls.registrationPostcodeFallbackUrl;
```
koepferd: As far as I understood in this case the `selectedAgency.url` would be used, right?
`registrationPostcodeFallbackUrl` is another specific use case in the scope of a consultingType and is only given when a consultingType doesn't have a white spot agency.

src/components/profile/AskerRegistrationExternalAgencyOverlay.tsx
koepferd: Yes, we still have to work on the general overlay styles for desktop: 
- icon left not center top
- buttons bottom right and next to each other

cypress/fixtures/service.consultingtypes.addiction.json
```diff
@@ -7,7 +7,9 @@
 		"welcome": "Willkommen bei der Suchtberatung!",
 		"registrationDropdown": "Suchtberatung"
 	},
-	"urls": {},
+	"urls": {
```
koepferd: See `AskerRegistrationExternalAgencyOverlay.tsx`, I think this is the wrong mock. This case should be tested for an agency not consultingType.

cypress/integration/profile.spec.ts
koepferd: What do you mean? Let's discuss on monday.

src/components/profile/AskerRegistrationExternalAgencyOverlay.tsx
koepferd: In addition, does it make sense to open the registration page of the agency in a new window instead of in the same tab in regards of usability? As a user I'd might check the offer of the agency and decide against it. Should be discussed with concept I think.

koepferd (changes requested):

src/components/profile/AskerRegistrationExternalAgencyOverlay.tsx
amannn: You're of course right, thanks for noticing this! Should be fixed now.

koepferd (dismissed):

koepferd (dismissed):

src/components/agencySelection/AgencySelection.tsx
amannn: `proposedAgencies` can be `null` during the registration. I'm currently investigating why this causes some tests on [the Diakonie side to fail](https://github.com/virtualidentityag/diakonie-onlineBeratung-frontend-theme/pull/9/checks?check_run_id=2943194125), but not here …

src/components/agencySelection/AgencySelection.tsx
```diff
@@ -82,6 +82,7 @@ export const AgencySelection = (props: AgencySelectionProps) => {
 	useEffect(() => {
 		if (isSelectedAgencyValidated()) {
 			const agency = {
+				...proposedAgencies?.find((cur) => cur.id === selectedAgencyId),
```
amannn, line 85: In the case of the registration outside of the app this can be undefined. We only need the `id` and `postcode` for that use case though and therefore this is sufficient.

koepferd (dismissed):

koepferd (approved):

PR #222, opened 2021-06-25
feat: 🎸 added ct dynamic texts for WelcomeScreen

koepferd:
## Proposed Changes

  - using texts for WelcomeScreen from CT-Service

- [x] wait for implementation in consultingType service

### DIFF
--- a/src/components/registration/Registration.tsx
+++ b/src/components/registration/Registration.tsx
@@ -84,6 +84,9 @@ export const Registration = ({ stageComponent: Stage }: RegistrationProps) => {
 							handleForwardToRegistration={
 								handleForwardToRegistration
 							}
+							welcomeScreenTexts={
+								consultingType.registration.welcomeScreen
+							}
 						/>
 					) : (
 						<RegistrationForm consultingType={consultingType} />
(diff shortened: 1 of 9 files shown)


iam-robin (dismissed):

src/components/registration/Registration.tsx
```diff
@@ -84,6 +84,9 @@ export const Registration = ({ stageComponent: Stage }: RegistrationProps) => {
 							handleForwardToRegistration={
 								handleForwardToRegistration
 							}
+							welcomeScreenTexts={
+								consultingType.registration.welcomeScreen
```
iam-robin: please use the global state instead of passing the texts as props

src/components/registration/WelcomeScreen.tsx
```diff
@@ -8,11 +8,13 @@ import { Button, ButtonItem, BUTTON_TYPES } from '../button/Button';
 import { Text } from '../text/Text';
 import { config } from '../../resources/scripts/config';
 import { Headline } from '../headline/Headline';
+import { RegistrationWelcomeScreenInterface } from '../../globalState';
 import './welcomeScreen.styles';
 
 interface WelcomeScreenProps {
 	title: string;
 	handleForwardToRegistration: Function;
+	welcomeScreenTexts: RegistrationWelcomeScreenInterface;
```
iam-robin: see other comment

src/components/registration/Registration.tsx
koepferd: Won't do, as we don't have a global state outside the app for now.
I changed the wording though to be more generic for future configurations.

iam-robin (approved):

PR #241, opened 2021-09-17
Feature optimize websocket

maeventer:
## Optimize performance

 - Prevent rocket chat websocket connection for all enquiries
 - Filter websocket events from technical rocket chat user to prevent multiple reloads of sessions

### DIFF
--- a/src/api/rocketChatSocket.ts
+++ b/src/api/rocketChatSocket.ts
@@ -164,8 +164,9 @@ export class rocketChatSocket {
 				response.fields.eventName === params[0];
 			if (changeResponseOnSubscribedEvent) {
 				const newMessage =
-					response.collection === SOCKET_COLLECTION.ROOM_MESSAGES &&
-					!response.fields.args[0].u.username.starsWith(
+					response.collection === SOCKET_COLLECTION.ROOM_MESSAGES;
+				const isTechnicalMessage =
+					response.fields.args[0].u.username.starsWith(
 						'cob-rocket-chat-technical-user'
 					);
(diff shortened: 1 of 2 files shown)


src/api/rocketChatSocket.ts
```diff
@@ -165,14 +165,18 @@ export class rocketChatSocket {
 			if (changeResponseOnSubscribedEvent) {
 				const newMessage =
 					response.collection === SOCKET_COLLECTION.ROOM_MESSAGES;
+				const isTechnicalMessage =
+					response.fields.args[0].u.username.startsWith(
+						'cob-rocket-chat-technical-user'
```
koepferd: Is this user on all systems named exactly the same?
I'd suggest a more generic name without 'cob' to reduce the client specifica. I am aware that you'd have to change the name of the technical user for that. Is the effort for that manageable?
I'd rather invest some time for this here as we can then use the generic name for the documentation and can be used by every instance of the onlineberatung and be less error-prone.
What do you think?

koepferd (changes requested):
See comment in rocketChatSocket.ts

src/api/rocketChatSocket.ts
maeventer: This expression matches for all systems hosted for caritas right now yes, but i agree with aspect having a unique generic technical username for all instances and would suggest having just the name 'rocket-chat-technical-user'. Changing the name is not that much effort so i would start adapt and document the changes ;-)

koepferd (approved):

PR #255, opened 2021-11-05
Hotfix teamsession lists

koepferd:
Fixes broken Routing for archive tabs


src/components/monitoring/Monitoring.tsx
```diff
@@ -101,7 +97,8 @@ export const Monitoring = () => {
 	const handleSubmit = () => {
 		apiUpdateMonitoring(activeSession.session.id, monitoringData)
 			.then((response) => {
-				handleBackButton();
+				const backLink: any = backLinkRef.current;
```
iam-robin: is the const necessary here?

koepferd: Nope, I removed it. Thanks!

iam-robin (approved):

PR #272, opened 2021-11-25
feat: [COBH-4010] Added delete option to chatFlyout menu

web-mi:
Fixes #COBH-4010

## Proposed Changes

  - Added delete option to chatFlyout menu for messages and archive
  - Fixed rerender of overlay buttons if type changed. (Effect hook was not triggered)

### DIFF
--- a/src/components/sessionMenu/SessionMenu.tsx
+++ b/src/components/sessionMenu/SessionMenu.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { useContext, useEffect, useState } from 'react';
+import { useCallback, useContext, useEffect, useState } from 'react';
 import { translate } from '../../utils/translate';
 import { config } from '../../resources/scripts/config';
 import { Link, Redirect, useLocation } from 'react-router-dom';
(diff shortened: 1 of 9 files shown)


src/resources/scripts/i18n/de/deleteSession.ts
```diff
@@ -0,0 +1,13 @@
+const deleteSession = {
+	'confirmOverlay.headline': 'Chat löschen',
+	'confirmOverlay.copy': 'Möchten Sie den Chat wirklich löschen?',
```
koepferd, line 3: Please also add corresponding informal texts.

src/components/sessionMenu/SessionMenu.tsx
```diff
@@ -516,6 +517,24 @@ export const SessionMenu = (props: SessionMenuProps) => {
 							{translate('chatFlyout.dearchive')}
 						</div>
 					)}
+				{!hasUserAuthority(AUTHORITIES.ASKER_DEFAULT, userData) &&
```
koepferd: Please check on: hasUserAuthority(AUTHORITIES.CONSULTANT_DEFAULT, userData).
This functionality should be available for all consultants.

There is also another asker authority, which we would need to exclude here aswell: ANONYMOUS_DEFAULT

koepferd:
After the success overlay is closed the open session list should be reloaded. I'll discuss this with you.

koepferd (changes requested):

PR #280, opened 2021-12-01
feat: [COBH-4055] allow escaping in markdownToDraft

web-mi:
Fixes #COBH-4055

## Proposed Changes

  - Allow escaping in markdownToDraft parsing
  - Custom markdown escaping logic for user input because build in method does not work for inline characters


koepferd (changes requested):
As discussed please check the functionality as it is buggy. Thank you!

maeventer (approved):

PR #285, opened 2021-12-07
fix: changed route component to render to prevent unmounting problems

web-mi:
Fixes #COBH-4068

## Proposed Changes

  - Changed route component attribute to render to prevent remounting of components
  - Removed useMemo from route rendering because its not required anymore
  - Changed some object deps to spreaded props to prevent triggering changes and forced reloads

### DIFF
--- a/src/components/app/Routing.tsx
+++ b/src/components/app/Routing.tsx
@@ -31,6 +31,8 @@ interface routingProps {
 export const Routing = (props: routingProps) => {
 	const { userData } = useContext(UserDataContext);
 	const { sessionsData } = useContext(SessionsDataContext);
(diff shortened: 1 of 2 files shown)


src/components/app/Routing.tsx
```diff
@@ -85,56 +86,46 @@ export const Routing = (props: routingProps) => {
...
-												type={route.type || null}
-											/>
-										)}
+					{routerConfig.detailRoutes.map(
+						(route: any): JSX.Element => (
+							<Route
+								exact
+								key={`detail-${route.path}`}
+								path={route.path}
+								test="test"
```
webatspeed: Couldn't find the reason for "test". Needed here?

webatspeed (approved):

PR #321, opened 2022-01-19
style: [DDO-9] made fallback notice more prominent

web-mi:
Fixes #DDO-9

## Proposed Changes

  - Added notice component
  - Changed styling of fallback notice


src/components/notice/Notice.tsx
```diff
@@ -0,0 +1,53 @@
...
+			case NOTICE_TYPE_INFO:
+			default:
+				return <InfoIcon />;
+		}
+	}, [type]);
+
+	return (
+		<div className={`notice notice--${type} ${className}`}>
+			<div className="notice__header">
+				<div className="notice__header__icon">{getIcon()}</div>
```
alxbenz: Is this an "approved" way of writing classnames, since I don't think this is correct BEM Syntax (I have seen this on a couple occasions in the project already)? Just wondering.

alxbenz (dismissed):

src/components/notice/Notice.tsx
koepferd: Hi @alxbenz, yes you are right, we have a technical debt here.
Thanks for bringing it up. I'd love if we can improve on this when touching old components.

koepferd (approved):

PR #346, opened 2022-02-03
Cobh 4140

alxbenz:
## Proposed Changes

  - check authority for asker to allow 2fa


koepferd (changes requested):
Actually you could just remove the whole check, as now every user can have 2fa and it's solely depending on the enable in the userData.

koepferd (approved):

PR #357, opened 2022-02-09
Cobh 4156 further steps

alxbenz:
- Added further steps for consultant to match last Message content

### DIFF
--- a/src/components/message/FurtherSteps.tsx
+++ b/src/components/message/FurtherSteps.tsx
@@ -24,7 +24,12 @@ import {
 	OVERLAY_FUNCTIONS
 } from '../overlay/Overlay';
 import { apiPutEmail, FETCH_ERRORS, X_REASON } from '../../api';
-import { ConsultingTypeInterface, UserDataContext } from '../../globalState';
+import {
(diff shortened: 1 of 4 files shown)


src/components/message/FurtherSteps.tsx
```diff
@@ -228,7 +247,7 @@ export const FurtherSteps = (props: FurtherStepsProps) => {
 							/>
 						</li>
 					</ul>
-					{showAddEmail && (
+					{!isConsultant && showAddEmail && (
```
koepferd, line 250: Please also add the condition to line 285, as the consultant shouldn't see this either.

In that edge case there won't be the illustration (example ct u25). I suggest to add the consultant info their aswell with an adjusted text like "Der_die Ratsuchende wurde über die nächsten Schritte informiert."

koepferd (changes requested):

src/components/message/FurtherSteps.tsx
koepferd: As discussed we won't have this case.

koepferd (approved):

PR #361, opened 2022-02-10
fix: [COBH-4159] fixed typing

web-mi:
Fixes #COBH-4159

## Proposed Changes

  - Moved typing logic to hook and simplified it
  - Fixed typing if activeSession isn't loaded before websocket is connected
  - Adding cancel "is typing" if message is send
  - Fixed send "is typing" if draft message is loaded


alxbenz (dismissed):
Code is looking good.
For the custom useTyping hook I would appreciate some comments since I can not really follow what each function does by just looking at the name. (though this is only relevant if the hook is intended to reusability).

alxbenz (approved):

PR #407, opened 2022-04-01
Obi 5 display name

alxbenz:


src/api/apiForwardMessage.ts
```diff
@@ -13,7 +13,7 @@ export const apiForwardMessage = async (
 	const data = JSON.stringify({
 		message: messageData,
 		timestamp: messageDate,
-		username: username,
+		username: displayName, // TODO change to displayName if message service is adjusted
```
koepferd, line 16: This can be updated now :-)
https://github.com/CaritasDeutschland/caritas-onlineBeratung-messageService/pull/78/files

alxbenz: I just talked to Torsten and username is still the required parameter. 
This change is rather to make sure the backend logs are not showing an error. 

So no, we can not change this to displayName currently.

koepferd: Ok, thank you for clarification!

src/resources/scripts/i18n/de/profileInformal.ts
```diff
@@ -24,6 +24,8 @@ const profileInformal = {
 		'Es ist ein Fehler aufgetreten. Bitte versuche es erneut.',
 	'data.personal.registrationLink.tooltip':
 		'Teile Deinen persönlichen Kontakt-Link mit jemanden, damit diese Person eine Onlineberatung direkt mit Dir starten kann.',
+	'data.info.public':
```
koepferd, line 27: Please use *Du to stay consistent

koepferd (changes requested):

koepferd (approved):

PR #420, opened 2022-04-20
[WIP] Feature/vic 226 core app walkthrough

koepferd:
## Proposed Changes

  - base functionality of product walkthrough


alxbenz:
SonarCloud is complaining about duplicated code. I had the same issue regarding the i18n files and excluded those from duplication detection. 
Maybe we should exclude this file also? src/components/walkthrough/steps.ts
Or move it to the i18n folder since it does not contain any application logic.

PR #439, opened 2022-05-16
fix: temporarily prevent double compile and hmr on webpack 5

ctmm:
https://github.com/facebook/create-react-app/issues/11961
https://github.com/webpack/webpack/issues/15431

Please note that this PR also introduces node 16 as a supported environment.
This is related but not strictly connected, so feel free to request it is done on a separate PR.


web-mi:
We fixed the performance issue in another pr.
Problem was sass-loader > 11 which has some problems with addionalData and recompiles all files on every change.

PR #441, opened 2022-05-19
Obi 267 - drag and drop for files inside chat container

mulkatz:
- drop container visual state changes according to drag position
  - inside drop zone
  - outside drop zone
  - outside browser window 
- stylings and animations


src/components/session/SessionItemComponent.tsx
```diff
@@ -1,21 +1,23 @@
 import * as React from 'react';
-import { useState, useContext, useEffect, useMemo } from 'react';
-import { useParams } from 'react-router-dom';
+import { useContext, useEffect, useMemo, useRef, useState } from 'react';
+import { Link, useLocation, useParams } from 'react-router-dom';
 import clsx from 'clsx';
 import {
```
mulkatz, line 5: It seems like prettier / organize import was not applied before this PR, so that import order changed in this file

alxbenz: It's not organizing on my side. So this must probably be something your IDE must do ;)
I like it though and maybe we could force it for everyone. What do you think? @web-mi @koepferd

src/components/dragAndDropArea/dragAndDropArea.styles.scss
```diff
@@ -0,0 +1,70 @@
...
+			.dropContainer__icon {
+				path {
+					fill: rgba(0, 0, 0, 0.87);
+				}
+			}
+		}
+	}
+}
+
+@supports (-webkit-backdrop-filter: none) or (backdrop-filter: none) {
```
alxbenz, line 64: We should probably configure stylelint to ignore IE 11 warnings...

src/components/dragAndDropArea/dragAndDropArea.styles.scss
```diff
@@ -0,0 +1,70 @@
...
+		right: 8px;
+		border-width: 2px;
+		border-style: dashed;
+		border-color: rgba(0, 0, 0, 0.4);
+		display: flex;
+		flex-direction: column;
+		justify-content: center;
+		align-items: center;
+		pointer-events: none;
+		border-radius: 15px;
```
alxbenz: Maybe we need to put some values as variables, so it is possible to change them in the theming repositories.

src/components/messageSubmitInterface/messageSubmitInterfaceComponent.tsx
```diff
@@ -248,6 +249,11 @@ export const MessageSubmitInterfaceComponent = (
 		};
 	}, []); // eslint-disable-line react-hooks/exhaustive-deps
 
+	useEffect(() => {
+		if (!props.preselectedFile) return;
+		handlePreselectedAttachmentChange();
+	}, [props.preselectedFile]); // eslint-disable-line react-hooks/exhaustive-deps
```
alxbenz: Please don't disable exhaustive-deps for new useEffects, since it might lead to unwanted side-effects.

We keep it for "old" code since we can not predict what will happen if we remove it, but we try to avoid it for new code.

src/components/session/SessionItemComponent.tsx
```diff
@@ -108,6 +116,11 @@ export const SessionItemComponent = (props: SessionItemProps) => {
 		}
 	};
 
+	useEffect(() => {
+		enableDraggingOnWindow();
+		return () => disableDraggingOnWindow();
+	}, []); // eslint-disable-line
```
alxbenz: [See the comment ](https://github.com/CaritasDeutschland/caritas-onlineBeratung-frontend/pull/441/files#r877102448)

src/resources/scripts/i18n/de/session.ts
```diff
@@ -19,7 +19,12 @@ const session = {
 	'groupChat.consultant.prefix': 'Moderator_in - ',
 	'monitoring.buttonLabel': 'Jetzt dokumentieren',
 	'u25.assignment.placeholder': 'Beratung zuweisen',
-	'unreadCount.maxValue': '99+'
+	'unreadCount.maxValue': '99+',
+	'dragAndDrop.explanation.insideDropArea':
```
alxbenz, line 23: We also have a file for informalLanguage. Please add the affected translations there as well.

src/components/dragAndDropArea/dragAndDropArea.styles.scss
alxbenz: #443

src/components/session/SessionItemComponent.tsx
mulkatz: It's a build-in feature on jetbrains IDEs, in VSCode it can be configured in the settings json with 
`"editor.codeActionsOnSave": {
    "source.organizeImports": true
},`

For an IDE independent solution, prettier-plugin-organize-imports should do the job

alxbenz: Can you add the prettier-plugin in a separate PR? (no rush)

src/components/messageSubmitInterface/messageSubmitInterfaceComponent.tsx
```diff
@@ -591,15 +603,18 @@ export const MessageSubmitInterfaceComponent = (
...
+		attachmentSizeMB > ATTACHMENT_MAX_SIZE_IN_MB
+			? handleLargeAttachments()
+			: displayAttachmentToUpload(attachment);
+	}, [handleLargeAttachments, props.preselectedFile]);
 
-	const handleLargeAttachments = () => {
-		removeSelectedAttachment();
-		setActiveInfo(INFO_TYPES.ATTACHMENT_SIZE_ERROR);
-	};
+	useEffect(() => {
```
mulkatz, line 614: @alxbenz If we do not suppress lint warnings for exhaustive-deps, we will lost the conventional order of hooks in functional components (hooks first, other functions after that), since properties in the deps array have to be declared before usage. This leads to useEffect hooks in the middle of components, which is a little annoying :/

alxbenz: Is this a thing?
Then I would prefer to disable the eslint rule for enforcing to declare the functions before the usage globally.

mulkatz: Ok, a deeper research has shown that this is not a thing in the end. So we should keep it like you suggested: lint-rules for exhaustive deps should not be disabled, instead useEffect should contain the complete deps array and use useCallback if needed.

@alxbenz I think the PR can be merged now

alxbenz (dismissed):

alxbenz (approved):

PR #448, opened 2022-06-01
Obi 403

MelanieJakobArbeit:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/src/components/profile/absenceFormular.styles.scss
+++ b/src/components/profile/absenceFormular.styles.scss
@@ -30,16 +30,16 @@
 		}
 
 		> p {
-			margin-bottom: 19px;
+			margin-bottom: $grid-base-two;
 		}
 
 		.mobile {
-			padding-top: 40px;
+			padding-top: $grid-base-five;
 		}
 
 		.disabled {
 			background-color: $white;
-			border: 1px solid rgba(0, 0, 0, 0.12);
+			border: 1px solid $black-10-opacity;
 		}
 
 		.textarea__wrapper {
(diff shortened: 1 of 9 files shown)


src/resources/scripts/i18n/de/profile.ts
```diff
@@ -4,6 +4,8 @@ const profile = {
 	'functions.title': 'Funktionen',
 	'functions.absenceTitle': 'Meine Abwesenheit',
 	'functions.absenceLabel': 'Hinterlegen Sie eine Abwesenheitsnachricht',
+	'functions.absenceActivatedLabel':
```
alxbenz, line 7: You can set the form of adress of the plattform to informal. For this purpose we also have informal language files. In this example there is also a profileInformal.ts.
You need to add this translation there as well and replace the "Sie" with "Du".

src/components/profile/absenceFormular.styles.scss
```diff
@@ -28,5 +28,22 @@
 				transition: all 0.3s ease;
 			}
 		}
+
+		> p {
+			margin-bottom: 19px;
```
alxbenz: in the settings.scss there are variables for pixel values ($grid-base-...). I think it would be better to use them where possible.

src/components/profile/absenceFormular.styles.scss
```diff
@@ -28,5 +28,22 @@
...
+		}
+
+		.mobile {
+			padding-top: 40px;
+		}
+
+		.disabled {
+			background-color: $white;
+			border: 1px solid rgba(0, 0, 0, 0.12);
```
alxbenz: Just an FYI: We often use variables for colors (in settings.scss) so they can be easily overwritten be a theming file in a theming repository. You probably don't need to use it here though, since I doubt this will be overwritten.

alxbenz (changes requested):

alxbenz (approved):

PR #456, opened 2022-06-10
feat: [OBI-459] re-encrypt room for users who lost their master key

mulkatz:
- re-encrypt room for users who lost their master key

### DIFF
--- a/src/hooks/useE2EE.tsx
+++ b/src/hooks/useE2EE.tsx
@@ -33,21 +33,22 @@ export const useE2EE = (rid: string): UseE2EEParams => {
 	const sessionKeyExportedStringRef = useRef(null);
 
 	// If hook is used search for members without key and set it
-	/*
-	ToDo: This is currenlty not working as fallback because room keys could not be resettet currenlty so there will never be users without keys
-	useEffect(() => {
+	const UpdateRoomKeysIfNecessary = useCallback(async () => {
 		if (!keyID || !sessionKeyExportedString || !rid) {
 			return;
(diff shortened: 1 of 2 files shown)


src/hooks/useE2EE.tsx
```diff
@@ -33,21 +33,22 @@ export const useE2EE = (rid: string): UseE2EEParams => {
 	const sessionKeyExportedStringRef = useRef(null);
 
 	// If hook is used search for members without key and set it
-	/*
-	ToDo: This is currenlty not working as fallback because room keys could not be resettet currenlty so there will never be users without keys
-	useEffect(() => {
+	const UpdateRoomKeysIfNecessary = useCallback(async () => {
```
alxbenz: can you use a smallcase update here?

src/hooks/useE2EE.tsx
```diff
@@ -61,10 +62,13 @@ export const useE2EE = (rid: string): UseE2EEParams => {
...
 					})
-			)
-		);
+			);
+		});
 	}, [keyID, rid, sessionKeyExportedString]);
-	 */
+
+	useEffect(() => {
+		UpdateRoomKeysIfNecessary();
```
alxbenz: see above

alxbenz (changes requested):

src/hooks/useE2EE.tsx
mulkatz: sorry, of course.. I've overlooked it

PR #457, opened 2022-06-10
feat: [OBI-148] optimize send attachment animation

mulkatz:
- optimize send attachment animation


alxbenz (approved):

alxbenz:
Since this is not relevant to the e2e / live listing, I would prefer this PR goes to develop

PR #467, opened 2022-06-12
fix: moved to sass-loader <= 10 & sass < 7

web-mi:
Fixes Webpack compile times

sass-loader > 10 with additionalData triggers recompile of all scss files on every change and makes webpack very slow.

## Proposed Changes

  - Moved to sass-loader 10.2.1
  - Moved node-sass to 6.0.1


alxbenz (dismissed):
Way better compilation times.

alxbenz (approved):

PR #470, opened 2022-06-14
Obi 369

MelanieJakobArbeit:
Fixes #

## Proposed Changes

  -
  -
  -


src/components/profile/ConsultantNotifications.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { translate } from '../../utils/translate';
+import { Headline } from '../headline/Headline';
+import { Text } from '../text/Text';
+import Switch from 'react-switch';
+import { UserDataContext } from '../../globalState';
+import { apiPatchUserData } from '../../api/apiPatchUserData';
+
+export const ConsultantNotifications = () => {
+	const { userData, setUserData } = useContext(UserDataContext);
+	const [isActivated, setIsActivated] = useState(false);
```
alxbenz: can we rename this variable, to represent the actual toggle it changes? Then we won't need to change it in the future if there are additional toggles added.

src/components/profile/ConsultantNotifications.tsx
```diff
@@ -0,0 +1,74 @@
...
+		updatedUserData.emailToggles.forEach((toggle) => {
+			if (
+				toggle.name === 'NEW_CHAT_MESSAGE_FROM_ADVICE_SEEKER' ||
+				toggle.name === 'NEW_FEEDBACK_MESSAGE_FROM_ADVICE_SEEKER'
+			) {
+				toggle.state = !toggle.state;
+			}
+		});
+		setIsActivated(!isActivated);
+		setUserData(updatedUserData);
```
alxbenz: Maybe wait until the patch user data call done to update the user data in case there is an error from the backend.

src/components/profile/ConsultantNotifications.tsx
```diff
@@ -0,0 +1,74 @@
...
+		apiPatchUserData(updatedUserData);
+	};
+
+	return (
+		<div className="notifivations">
```
alxbenz: typo

src/components/profile/ConsultantNotifications.tsx
```diff
@@ -0,0 +1,74 @@
...
+					text={translate('profile.notifications.title')}
+					semanticLevel="5"
+				/>
+				<Text
+					text={translate('profile.notifications.subtitle')}
+					type="standard"
+					className="tertiary"
+				/>
+			</div>
+			{userData.emailToggles.length > 0 && (
```
alxbenz: Does it make sense to show the headline and subline if there is no toggle?

alxbenz (changes requested):

src/components/profile/ConsultantNotifications.tsx
web-mi: I think we don't need this state anyway becuase all data alredy in userData.emailToggles.
Just map the toggles.
On click send the new values.
Update UserData by useUpdateUserData hook.

Maybe keep a clone of userData.emailToggles in a state so the toggle changes before requests are finished.

web-mi (changes requested):

src/components/profile/ConsultantNotifications.tsx
alxbenz: Even better ;)

src/components/profile/ConsultantNotifications.tsx
```diff
@@ -0,0 +1,84 @@
...
+						/>
+						<Text
+							text={translate('profile.notifications.subtitle')}
+							type="standard"
+							className="tertiary"
+						/>
+					</div>
+
+					<div className="flex">
+						<Switch
```
web-mi, line 48: userData.emailToggles.map((emailToggle) => ...)

src/components/profile/ConsultantNotifications.tsx
```diff
@@ -0,0 +1,84 @@
...
+							className="mr--1"
+							onChange={() => toogleSwitch()}
+							checked={isFollowUpMessagesActivated}
```
web-mi: checked={emailToggle.state}

src/components/profile/ConsultantNotifications.tsx
```diff
@@ -0,0 +1,84 @@
...
+							width={48}
+							height={26}
+							onColor="#0dcd21"
+							offColor="#8C878C"
+							boxShadow="0px 1px 4px rgba(0, 0, 0, 0.6)"
+							handleDiameter={27}
+							activeBoxShadow="none"
+						/>
+						<Text
+							text={translate('profile.notifications.label')}
```
web-mi: text={translate(`profile.notifications.label.${emailToggle.name}`)}

web-mi (dismissed):

src/resources/scripts/config.ts
```diff
@@ -110,3 +112,15 @@ export const config = {
 		enableEncodedTransformSupport: false
 	}
 };
+
+export const emails = {
```
alxbenz, line 116: I am not sure if we should put this in the global config or if it would be better inside the Notifications component, since it is not used anywhere else. What do you think?

alxbenz (approved):

alxbenz (approved):

PR #489, opened 2022-06-28
OBI-431

mulkatz:
- handle master key lost scenario
  - show info message + overlay
  - send / receive special system alias message
  - find out ability to send info message


src/components/message/MessageItemComponent.tsx
```diff
@@ -262,6 +267,36 @@ export const MessageItemComponent = ({
...
+						/>
+
+						<div
+							className={clsx(
+								'messageItem__message',
+								isMyMessage && 'messageItem__message--myMessage'
+							)}
+						>
+							<span
+								dangerouslySetInnerHTML={{
```
alxbenz: Does this need to be set as HTML?
Wouldn't <span>{translate(...)}</span> suffice here?

src/components/session/SessionItemComponent.tsx
```diff
@@ -450,6 +456,50 @@ export const SessionItemComponent = (props: SessionItemProps) => {
...
+	const sendMessageLostInfo = async () => {
+		await apiSendAliasMessage({
+			rcGroupId: groupIdFromParam,
+			type: ALIAS_MESSAGE_TYPES.MASTER_KEY_LOST
+		});
+	};
+
+	const canDecrypted = keyID !== null;
+
+	// const canDecrypted = useCallback(() => {
```
alxbenz: Can we remove this?

alxbenz:
I changed some minor stuff.
But during my testing this was not working as intended... Probably because this was referencing and older version of e2ee (since then we change the useE2ee hook to not trigger in all the list items, as well as changed the subscriptions to use the websocket instead of the API).
@Franjoo can you have a look again?
I could not send an Alias Message after losing the masterKey. And the other user was not able to reset my subscription key.
Feel free to contact me next week for a quick chat.

mulkatz:
> I changed some minor stuff. But during my testing this was not working as intended... Probably because this was referencing and older version of e2ee (since then we change the useE2ee hook to not trigger in all the list items, as well as changed the subscriptions to use the websocket instead of the API). @Franjoo can you have a look again? I could not send an Alias Message after losing the masterKey. And the other user was not able to reset my subscription key. Feel free to contact me next week for a quick chat.

The problem why the update group key mechanic not worked as intended anymore was caused by a more recent change in the addNewUsersToEncryptedRoom function. More precisely:

```
const { users } = await apiRocketChatGetUsersOfRoomWithoutKey(rid);

// we can stop this request if there are only system and technical user without keys
if (users.length <= 2) { // <- that was always true in case of master key lost. setting 2 to 1 solved the issue and got it working again  
  return;
}
```

Besides from that, I noticed your changes included the introduction of the possibility to send unencrypted messages while normally sending encrypted ones. I used your new login and changed the message type when sending the master key lost info. Now an unencrypted user message is sent instead of a system message. That has the advantage, that is is clear which participant has send the info message in the past. (That would be unclear if both participant have a decrypted chat / present master keys; the participant with a decrypted chat would think that the other could not decrypt anymore). 
If thats not entirely clear, lets have a chat after your vacation ;)

src/resources/styles/settings.scss
```diff
@@ -206,6 +206,14 @@ $message-background: #d64b49;
 $message-submit-interface-textarea-background-color: $dark-grey;
 $message-submit-interface-textarea-background-color-yellow: $yellow-3;
 $message-submit-interface-textarea-placeholder-color: $light-grey;
+$message-submit-interface-icon-disabled: #cac8c6;
+$message-submit-interface-textarea-background-color-disabled: #f0eeed;
+$message-submit-interface-encryption-notice-info-wrapper-disabled: rgba(
```
web-mi: Oben bitte 
$black-5-opacity: rgba(0, 0, 0, 0.5);
definieren und hier verwenden.
Dann bitte auch gleich bei $tab-background-color nutzen

src/resources/scripts/i18n/de/session.ts
```diff
@@ -33,7 +33,26 @@ const session = {
 		'Legen Sie die Datei hier ab, um sie hochzuladen.',
 	'dragAndDrop.explanation.outsideDropArea':
 		'Ziehen Sie die Datei in das Feld, um sie hochzuladen.',
-	'dragAndDrop.restrictions': '.jpg, .png, .pdf, .docx, .xlsx bis maximal 5MB'
+	'dragAndDrop.restrictions':
+		'.jpg, .png, .pdf, .docx, .xlsx bis maximal 5MB',
+	'encrypted.notice.send.info':
+		'Ich kann unsere Nachrichten aus technischen Gr&uuml;nden nicht mehr lesen.',
+	'encrypted.notice.send.first':
```
web-mi: first und second kannst du in einen Satz mit einem Platzhalter packen den du dann beim translate(..., {platzhalter: 'xy'}) übergeben kannst. 
Siehe qrCode.ts -> download.filename

src/resources/scripts/i18n/de/sessionInformal.ts
```diff
@@ -13,7 +13,20 @@ const sessionInformal = {
 	'dragAndDrop.explanation.insideDropArea':
 		'Lege die Datei hier ab, um sie hochzuladen.',
 	'dragAndDrop.explanation.outsideDropArea':
-		'Ziehe die Datei in das Feld, um sie hochzuladen.'
+		'Ziehe die Datei in das Feld, um sie hochzuladen.',
+	'encrypted.notice.send.first':
```
web-mi: Same like session.ts. 
first und second durch platzhalter ersetzten

src/resources/scripts/i18n/de/sessionInformal.ts
```diff
@@ -13,7 +13,20 @@ const sessionInformal = {
...
+		'Sende die untenstehende Nachricht und benachrichtige ',
+	'encrypted.notice.send.second':
+		', dass Du aus technischen Gründen keine Nachrichten lesen kannst.',
+	'encrypted.notice.first':
+		'Aus technischen Gründen kannst Du keine Nachrichten lesen und schreiben. Wenn ',
+	'encrypted.notice.second':
+		' den Chat öffnet, kannst Du die Nachrichten wieder lesen und neue Nachrichten schreiben.',
+	'encrypted.overlay.text.first':
+		'Deine Nachrichten sind aus Sicherheitsgründen Ende-zu-Ende verschlüsselt. Das bedeutet, niemand außerhalb dieses Chats kann die Nachrichten lesen. Nicht einmal die Online-Beratungs-Plattform. <br/><br/>Da Du dein Passwort zurückgesetzt hast, sind die Nachrichten für dich momentan nicht lesbar. Sobald ',
```
web-mi:
```suggestion
		'Deine Nachrichten sind aus Sicherheitsgründen Ende-zu-Ende verschlüsselt. Das bedeutet, niemand außerhalb dieses Chats kann die Nachrichten lesen. Nicht einmal die Online-Beratungs-Plattform. <br/><br/>Da Du Dein Passwort zurückgesetzt hast, sind die Nachrichten für Dich momentan nicht lesbar. Sobald ',
```

web-mi (changes requested):

src/resources/scripts/i18n/de/session.ts
```diff
@@ -33,7 +33,26 @@ const session = {
...
+		'Aus technischen Gründen können Sie keine Nachrichten lesen und schreiben. Wenn ',
+	'encrypted.notice.second':
+		' den Chat öffnet, können Sie die Nachrichten wieder lesen und neue Nachrichten schreiben.',
+	'encrypted.notice.more': ' Mehr erfahren',
+	'encrypted.overlay.headline': 'Ende-zu-Ende Verschlüsselung',
+	'encrypted.overlay.text.first':
+		'Ihre Nachrichten sind aus Sicherheitsgründen Ende-zu-Ende verschlüsselt. Das bedeutet, niemand außerhalb dieses Chats kann die Nachrichten lesen. Nicht einmal die Online-Beratungs-Plattform. <br/><br/>Da Sie Ihr Passwort zurückgesetzt haben, sind die Nachrichten für Sie momentan nicht lesbar. Sobald ',
+	'encrypted.overlay.text.second':
+		' den Chat öffnet, können Sie die Nachrichten wieder lesen und auch schreiben.',
+	'encrypted.overlay.button': 'Schliessen',
```
web-mi:
```suggestion
	'encrypted.overlay.button': 'Schließen',
```
Bitte überall gleich ersetzten. Wir haben aktuell einen Mix aus Schliessen und Schließen

web-mi (changes requested):

PR #507, opened 2022-07-07
fix: [OBI-588] remove call to tenant service if feature flag is disab…

mulkatz:
remove call to tenant service if feature flag is disabled


src/utils/useTenantTheming.ts
```diff
@@ -221,10 +221,7 @@ const useTenantTheming = () => {
 
 	useEffect(() => {
 		if (!subdomain || !config.enableTenantTheming) {
-			apiGetTenantTheming({ subdomain }).then(({ settings }) => {
```
mulkatz, line 224: @CarlosSoares I don't understand that line :/ Why doing the call if feature flag is disabled or got not subdomain param?

CarlosSoares: We've a new feature flag that is stored in the tenant settings 
image

So that was requested to have that feature toggle, so for that we need to do the request to check the value

CarlosSoares: @Franjoo if you notice I only get the `settings` object, not the theming

mulkatz: @CarlosSoares Ok, got it, but this cause an error to occur in caritas since a there a tenant service is not present. I should have solved that by using an frontend feature flag that can disable the request against the tenant service.

CarlosSoares: I didn't know that we didn't have the tenant service in the caritas, then maybe the calls to tenant service shouldn't be here ? 🤔

CarlosSoares (approved):

src/utils/useTenantTheming.ts
mulkatz: Ok, we do not call the tenant service anymore if local feature flag is disabled

PR #512, opened 2022-07-12
feat: when auto select post code and the feature flag is enabled 

CarlosSoares:
Fixes #
When we've the autoSelectPostalCode at true we need to set the preselected agency

### DIFF
--- a/src/components/registration/RegistrationForm.tsx
+++ b/src/components/registration/RegistrationForm.tsx
@@ -24,7 +24,8 @@ import {
 	AgencyDataInterface,
 	ConsultantDataInterface,
 	ConsultingTypeInterface,
-	LegalLinkInterface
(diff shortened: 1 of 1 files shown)


src/components/registration/RegistrationForm.tsx
```diff
@@ -110,6 +115,30 @@ export const RegistrationForm = ({
 		}
 	}, [formAccordionValid, isDataProtectionSelected]);
 
+	useEffect(() => {
+		const shouldRequest =
```
koepferd: Can you please change the naming to something more meaningful to have it clear on the first look (as in what/why requested)?

koepferd (changes requested):

koepferd (approved):

PR #555, opened 2022-08-02
fix: only show Erstanfragen chats on Erstanfragen page

tarikalmashni:


src/components/sessionsList/SessionsList.tsx
```diff
@@ -694,6 +694,9 @@ export const SessionsList = ({
 				// filter teamsessions only without my user id as consultant
 				case SESSION_LIST_TYPES.TEAMSESSION:
 					return session?.consultant?.id !== userData.userId;
+				// only show sessions without an assigned consultant in sessionPreview
+				case SESSION_LIST_TYPES.ENQUIRY:
+					return session?.consultant === null;
```
CarlosSoares:
```suggestion
					return !session?.consultant;
```

this way it also works if the consultant is undefined

CarlosSoares (approved):

PR #556, opened 2022-08-02
feat(obi-383): changed 2fa dates logic

alxbenz:
Fixes #

## Proposed Changes

  -
  -
  -


web-mi (approved):

koepferd:
Hi @web-mi , @alxbenz ,
in my opinion the fixed dates shouldn't be in the core code as this will trigger the functionality in every instance.
Default it should be deactivated.

PR #570, opened 2022-08-05
feat: add the copy links (qr code and link) to group chages info

CarlosSoares:
Adding the copy links to group chat info to allow us login and join right away to the chat

**Still missing the feature toggle and one text(check the translation: copy.link.notification.text) that I've asked Lea to provide**

image


tarikalmashni (dismissed):

tarikalmashni (approved):
Talked with Carlos about it. Only the connection to the feature toggle is missing (currently represented by a local variable which defaults to **false**).

PR #575, opened 2022-08-08
fix: issue with zero agency inside conditional booking

Soarecos:


src/components/booking/booking.tsx
```diff
@@ -40,7 +40,7 @@ export const Booking = () => {
 			getCounselorAppointmentLink(consultantId).then((response) => {
 				setAppointmentLink(response.slug);
 			});
-		} else if (agencyId) {
+		} else if (agencyId >= 0) {
```
adnanalicic: you can do it like this, or you can replace else if with else, since agencyId must always be provided in this case.

adnanalicic (dismissed):

adnanalicic (approved):

PR #588, opened 2022-08-12
feat: changes from vi-saas-frontend updated in core app

sergiofv:
changes from vi-saas-frontend remade on the core app, so we don't have commits behind in our vi-saas-frontend repo, and the code is actually in the repo where it should be.


web-mi (dismissed):

tarikalmashni (dismissed):

src/resources/styles/settings.scss
```diff
@@ -266,3 +266,5 @@ $border-radius-e2ee-hint: 4px;
 $border-color-e2ee-hint: #4fcc5c;
 $background-color-e2ee-hint: #caf0ce;
 $drag-and-drop-area-border-radius: $grid-base-two;
+$input-width: 320px;
```
koepferd: please use more specific variable names for which component this is used.

koepferd (changes requested):

src/resources/styles/settings.scss
sergiofv: I renamed it to $chat-input-width

tarikalmashni (approved):

koepferd (approved):

PR #595, opened 2022-08-22
feat: connect frontend to group chat api v2

tarikalmashni:
Enables the creation of V2 group chats that have no consultingType
This PR is only for testing and not final as V2 is auto-enabled while creating group chats (instead of relying on a feature flag).

### DIFF
--- a/src/components/sessionsListItem/SessionListItemComponent.tsx
+++ b/src/components/sessionsListItem/SessionListItemComponent.tsx
@@ -200,7 +200,7 @@ export const SessionListItemComponent = ({
 				>
 					<div className="sessionsListItem__row">
 						<div className="sessionsListItem__consultingType">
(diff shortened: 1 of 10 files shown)


src/components/sessionsListItem/SessionListItemComponent.tsx
```diff
@@ -200,7 +200,7 @@ export const SessionListItemComponent = ({
 				>
 					<div className="sessionsListItem__row">
 						<div className="sessionsListItem__consultingType">
-							{consultingType.titles.default}
+							{consultingType?.titles?.default}
```
CarlosSoares, line 203: create a const with this value that way you can reuse on the line 298

CarlosSoares (approved):
Left you a small comment 💅

PR #623, opened 2022-09-12
feat: adding the new api to retrieve the settings from the server

CarlosSoares:
## Proposed Changes

  - Adding the new way of having the app settings in an api instead of having static in the config file
  - This is protected by a feature flag called **useApiClusterSettings** so if this is enabled then we use the api settings otherwise we use the settings from the config file.
  - By default even with the flag **useApiClusterSettings** enabled we still apply the defaults from the config file
  - **Breaking change**: From now on when we want to check something regarding settings we need to use the hook **useAppConfigContext** this handles all the settings.
  - The new hook is prepared to adapt to new cluster toggles, that means that we just need to add the new key to the interface called **AppConfigInterface** and that's it, if you need some default value is also needed to set the default value in the function **UseAppConfigProvider**

### DIFF
--- a/src/globalState/context/useAppConfig.tsx
+++ b/src/globalState/context/useAppConfig.tsx
@@ -1,0 +1,78 @@
+import React, { useCallback } from 'react';
+import { config } from '../../resources/scripts/config';
+import { AppConfigInterface } from '../interfaces/AppConfigInterface';
+import { ServerAppConfigInterface } from '../interfaces/ServerAppConfigInterface';
+
+interface AppConfigContextInterface {
+	settings: AppConfigInterface;
+	setServerSettings: (settings: ServerAppConfigInterface) => void;
+	setManualSettings: (settings: Partial<AppConfigInterface>) => void;
+}
+
+const UseAppConfigContext =
+	React.createContext<
+		[
+			AppConfigInterface,
+			React.Dispatch<React.SetStateAction<AppConfigInterface>>
+		]
+	>(null);
+
+const UseAppConfigProvider = ({
+	children
+}: {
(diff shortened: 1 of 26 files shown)


src/globalState/context/useAppConfig.tsx
```diff
@@ -0,0 +1,78 @@
...
+					...current,
+					[key]: serverSettings[key].value
+				}),
+				{} as Record<string, boolean>
+			);
+			setNewSettings({
+				...settings,
+				...(finalServerSettings as unknown as AppConfigInterface)
+			});
+			console.log({ ...settings, ...finalServerSettings });
```
Soarecos: Maybe we should remove the console.log

CarlosSoares: Ups 😃 
Removed

Soarecos (dismissed):

src/resources/scripts/config.ts
```diff
@@ -21,6 +21,8 @@ export const config = {
 	disableVideoAppointments: false, // Feature flag to enable Video-Termine page
 	useMultiTenancyWithSingleDomain: false, // Feature flag to enable the multi tenancy with a single domain ex: lands
 	useTenantService: false,
+	useApiClusterSettings: true, // Feature flag to enable the cluster use the cluster settings instead of the config file
```
koepferd: @CarlosSoares, please disable per default as the current open source version does not include the k8s setup yet.

koepferd (changes requested):

src/resources/scripts/config.ts
CarlosSoares: Done

koepferd (approved):

PR #634, opened 2022-09-16
fix: budibase login location

Soarecos:


src/components/app/AuthenticatedApp.tsx
```diff
@@ -41,13 +43,34 @@ export const AuthenticatedApp = ({
 	const { settings, setServerSettings } = useAppConfigContext();
 	const { setConsultingTypes } = useContext(ConsultingTypesContext);
 	const { userData, setUserData } = useContext(UserDataContext);
+	const { tenant } = useContext(TenantContext);
```
CarlosSoares: you can use the `useTenant` hook instead

src/components/app/AuthenticatedApp.tsx
```diff
@@ -41,13 +43,34 @@ export const AuthenticatedApp = ({
...
 
 	const [appReady, setAppReady] = useState<boolean>(false);
 	const [loading, setLoading] = useState<boolean>(true);
 	const [userDataRequested, setUserDataRequested] = useState<boolean>(false);
 
 	const { notifications } = useContext(NotificationsContext);
 
+	const loginBudiBase = useCallback(() => {
```
CarlosSoares: couldn't this be moved into a helper/hook? 

On the helper something like this:
```
export const useLoginBudiBase = () => {
  const tenant = useTenant();
  const login = useCallback(() => {
    iframe code
  }, []);
  
  return { login };
};
```
On this file
```
const tenant = useTenant();
const { login } = useLoginBudiBase();
useEffect(() => {
    if (tenant?.settings?.featureToolsEnabled) {
  	login();
    }
  }, [login, tenant?.settings?.featureToolsEnabled]);

```

CarlosSoares (changes requested):

src/components/app/AuthenticatedApp.tsx
Soarecos: I just didn't get it the benefit of using this hook instead of tenantContext, could you explain pls?

CarlosSoares (approved):

PR #666, opened 2022-09-27
fix: update menu on booking event

Soarecos:


sergiofv (dismissed):

src/components/booking/hooks/useAskerHasAssignedConsultant.ts
```diff
@@ -13,14 +14,15 @@ export const useAskerHasAssignedConsultant = () => {
...
+	const { sessions } = useContext(SessionsDataContext);
 
 	useEffect(() => {
 		if (isAdviceSeeker) {
 			apiGetAskerSessionList().then(({ sessions }) => {
 				setAssignedConsultant(!!sessions?.[0]?.consultant);
 			});
 		}
-	}, [userData, isAdviceSeeker]);
+	}, [userData, isAdviceSeeker, sessions]);
```
CarlosSoares: are you sure of this? What has the reason behind ?

Soarecos: On the first time booking is needed to updated the navigation to appear the "Meine Termine" menu. Otherwise we would need to refresh the page. This was the approach I found, but pls fell free to suggest another :)

CarlosSoares (approved):

PR #673, opened 2022-09-30
feat: allow user join a group chat when has a valid link

CarlosSoares:
## Proposed Changes

  - Allow join to a group chat when the feature group chat v2 is enabled
  - Remove the termine buttons when we're in a group chat


src/components/groupChat/GroupChatInfo.tsx
```diff
@@ -71,6 +71,7 @@ export const GroupChatInfo = () => {
 	const [redirectToSessionsList, setRedirectToSessionsList] = useState(false);
 	const [isRequestInProgress, setIsRequestInProgress] = useState(false);
 	const [bannedUsers, setBannedUsers] = useState<string[]>([]);
+	const [isV2GroupChat, setIsV2GroupChat] = useState<boolean>(false);
```
Soarecos, line 74: It's needed to say it's a boolean even when the default value is already "false"?

Soarecos (approved):

sergiofv (approved):

PR #691, opened 2022-10-07
fix: budibase logout call

Soarecos:


src/components/login/Login.tsx
```diff
@@ -116,8 +116,11 @@ export const Login = ({ legalLinks, stageComponent: Stage }: LoginProps) => {
 	useEffect(() => {
 		setOtp('');
 		setIsOtpRequired(false);
+	}, [username]);
```
CarlosSoares, line 119: why this needs the username?

CarlosSoares (approved):

PR #742, opened 2022-11-09
fix: wrong time shown in appointment messages

adnanalicic:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/src/containers/bookings/components/Calcom/Cal.tsx
+++ b/src/containers/bookings/components/Calcom/Cal.tsx
@@ -70,7 +70,9 @@ export default function Cal({
 
 					const sessionId = sessions[0]?.session?.id;
 					const { data } = e.detail;
-					const date = data.date;
+					const date = new Date(data.date)
+						.toUTCString()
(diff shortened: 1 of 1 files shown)


Soarecos (approved):

src/containers/bookings/components/Calcom/Cal.tsx
```diff
@@ -70,7 +70,9 @@ export default function Cal({
 
 					const sessionId = sessions[0]?.session?.id;
 					const { data } = e.detail;
-					const date = data.date;
+					const date = new Date(data.date)
+						.toUTCString()
```
Soarecos, line 74: It's ok be "UTC"?

web-mi: We have a function for prettyDate if you need to format the date for the ui

adnanalicic: it's ok like this. its a calcom bug and unfortunately not used at all on frontend... long story

PR #756, opened 2022-11-15
feat: VIC-1759 appointment notes on chat

Soarecos:


src/components/message/Appointment.tsx
```diff
@@ -106,21 +110,58 @@ export const Appointment = (param: {
 						className="appointmentSet__time"
 					/>
 				</div>
-				<div className="appointmentSet--flex">
```
CarlosSoares, line 109: This is ok removing this block?

src/components/message/Appointment.tsx
```diff
@@ -106,21 +108,53 @@ export const Appointment = (param: {
...
+						{parsedData.note.length > 110 && (
+							<>
+								<div
+									className="appointmentSet__note appointmentSet--flex appointmentSet--pointer"
+									onClick={() => setExpanded(!expanded)}
+								>
+									<Text
+										text={
+											expanded
+												? translate(
```
CarlosSoares: maybe create a const for this translation key? 
```
translate(translateKey);
```

src/components/message/Appointment.tsx
Soarecos: for now yes, we gonna refactor in the future :)

CarlosSoares (approved):

PR #765, opened 2022-11-21
fix: remove consulting type when topic are enabled

CarlosSoares:
When the topics are disabled and the auto Select post code is disabled
image
When the topics are disabled and the auto Select post code is enabled
image
When the topics are enabled and the auto Select post code is enabled
image
When the topics are enabled and the auto Select post code is disabled
image


src/globalState/provider/UserDataProvider.tsx
```diff
@@ -18,3 +18,7 @@ export function UserDataProvider(props) {
 		</UserDataContext.Provider>
 	);
 }
+
+export const useUserData = (): UserDataInterface => {
+	return React.useContext(UserDataContext)?.userData;
```
Soarecos: This doesn't seems to be used anywhere

Soarecos (dismissed):

Soarecos (approved):

PR #789, opened 2022-12-05
feat(obi-873): hide decorative elements for screen readers

RunzelRosinchen:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/src/components/overlay/Overlay.tsx
+++ b/src/components/overlay/Overlay.tsx
@@ -246,7 +246,11 @@ const OverlayContent: VFC<Omit<OverlayProps, 'name'>> = (props) => {
 											: ''
 									}`}
 								>
-									<Illustration />
+									<Illustration
+										aria-hidden="true"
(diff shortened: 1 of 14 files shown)


src/components/overlay/Overlay.tsx
```diff
@@ -246,7 +246,11 @@ const OverlayContent: VFC<Omit<OverlayProps, 'name'>> = (props) => {
 											: ''
 									}`}
 								>
-									<Illustration />
+									<Illustration
+										aria-hidden="true"
+										focusable="false"
+										title=""
```
alxbenz, line 252: No title for this illustration?

alxbenz (approved):

src/components/overlay/Overlay.tsx
RunzelRosinchen: Some of those svgs have a "_" as title in there files that shouldn' show up here

PR #798, opened 2022-12-08
Feat/vic 2126 accept changed tc and privacy policy

adnanalicic:
Fixes #

## Proposed Changes

  -
  -
  -


src/components/termsandconditions/TermsAndConditions.tsx
```diff
@@ -0,0 +1,242 @@
...
+	const transformText2DataPrivacyLink = (text: string) => {
+		let hereLabel = translate('termsAndConditionOverlay.labels.here');
+		return text.replace(
+			hereLabel,
+			`<a class='link' href='/datenschutz'>${hereLabel}</a>`
+		);
+	};
+
+	useEffect(() => {
+		const standardButtons = (userConfirmed: boolean) => {
```
CarlosSoares: this function can be outside of the useEffect

src/components/termsandconditions/TermsAndConditions.tsx
```diff
@@ -0,0 +1,242 @@
...
+				},
+				{
+					label: translate('termsAndConditionOverlay.buttons.accept'),
+					disabled: !userConfirmed,
+					type: BUTTON_TYPES.PRIMARY
+				}
+			];
+		};
+
+		const dataPrivacyButtons = [
```
CarlosSoares: same as this const

src/components/termsandconditions/TermsAndConditions.tsx
```diff
@@ -0,0 +1,242 @@
...
+			`<a class='link' href='/nutzungsbedingungen'>${termsLabel}</a>`
+		);
+		return text.replace(
+			privacyLabel,
+			`<a class='link' href='/datenschutz'>${privacyLabel}</a>`
+		);
+	};
+
+	const transformText2DataPrivacyLink = (text: string) => {
+		let hereLabel = translate('termsAndConditionOverlay.labels.here');
```
CarlosSoares, line 62: why you need this type of transforms?

src/components/termsandconditions/TermsAndConditions.tsx
```diff
@@ -0,0 +1,242 @@
...
+				showOverlay: true,
+				userConfirmed: viewState.userConfirmed,
+				buttons: dataPrivacyButtons
+			});
+		}
+		// eslint-disable-next-line react-hooks/exhaustive-deps
+	}, [viewState.userConfirmed]);
+
+	if (!viewState.showOverlay) {
+		return <></>;
```
CarlosSoares: `return null` instead of `return <></>`

src/components/termsandconditions/TermsAndConditions.tsx
```diff
@@ -0,0 +1,242 @@
...
+		return (
+			userData.termsAndConditionsConfirmation === null ||
+			new Date(userData.termsAndConditionsConfirmation) <
+				new Date(tenantData.content.termsAndConditionsConfirmation)
+		);
+	}
+	return false;
+};
+
+const privacyChanged = (
```
CarlosSoares: you could create a generic function instead of replicated only because of the fields name changed :)

you can do something like:
```

const hasChanged = (
	tenantData: TenantDataInterface,
	userData: UserDataInterface,
	field: string,
) => {
	if (tenantData) {
		return (
			userData.termsAndConditionsConfirmation === null ||
			new Date(userData[field]) <
			new Date(tenantData.content[field])
		);
	}
	return false;
}
```

src/components/termsandconditions/TermsAndConditions.tsx
adnanalicic: done

src/components/termsandconditions/TermsAndConditions.tsx
adnanalicic: done

src/components/termsandconditions/TermsAndConditions.tsx
adnanalicic: some words in text contain links

src/components/termsandconditions/TermsAndConditions.tsx
adnanalicic: done

src/components/termsandconditions/TermsAndConditions.tsx
CarlosSoares: In that case you should use the tool to do this
https://www.i18next.com/translation-function/interpolation

src/components/termsandconditions/TermsAndConditions.tsx
adnanalicic: done

CarlosSoares (approved):

PR #803, opened 2022-12-12
fix: adding the initial cookie language

CarlosSoares:
Fixes #
Add the initial cookie language value when the apps starts


Soarecos (dismissed):

web-mi:
Looks good. Could you move your adaption in the LocaleSwitch to the second useEffect too and remove it from the localeSwtich?

web-mi (approved):

PR #894, opened 2023-04-11
fix(styles): remove duplicate styles

SheepFromHeaven:
Fixes #

## Proposed Changes

  -
  -
  -

### DIFF
--- a/src/components/stage/stage.styles.scss
+++ b/src/components/stage/stage.styles.scss
@@ -132,7 +132,6 @@ $gridSpacing: $grid-base-four;
 	}
 
 	&__spinner {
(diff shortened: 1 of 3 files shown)


src/components/stage/stage.styles.scss
```diff
@@ -132,7 +132,6 @@ $gridSpacing: $grid-base-four;
 	}
 
 	&__spinner {
-		animation-delay: 2s;
```
CarlosSoares, line 135: Are you sure about this one?

SheepFromHeaven: oops, you are right! fixed

PR #946, opened 2023-05-31
fix: login after password reset OB-5109

CarlosSoares:
The issue occurs when the user has his password defined then he resets the password and change it to a different one and then the error happens


vi-sudo:
Task linked: [OB-5109 First attempt login issue directly after reset password](https://app.clickup.com/t/6656196/OB-5109)

tkuzynow (approved):

PR #961, opened 2023-06-21
Fix some issues found in group chat

CarlosSoares:
Fixes #

## Proposed Changes

  -
  -
  -


src/resources/i18n/de.json
```diff
@@ -695,6 +695,17 @@
 		}
 	},
 	"groupChat": {
+		"qrCode": {
```
RunzelRosinchen, line 698: Don't you need to add that as well to en and de.informal?

RunzelRosinchen (changes requested):

RunzelRosinchen (approved):

PR #1024, opened 2024-03-13
fix: removed default consultingType overrides

web-mi:
Fixes #

## Proposed Changes

  -
  -
  -


janrembold (approved):
So mag ich das!!! +2 -904 loc 🥇

